PR #57, opened 2016-01-27
Allow tasks to be filtered by status, Mark an item as undone

joehandlebar:
Users can filter tasks by status. The status filter is applied after tasks are filtered by owner.
Users can toggle task status between 'done' and 'undone'.

- there's a new pipe called ```status```
- removed ```ngc-task-count```, created ```ngc-task-filters```

Help:

- is there a TypeScript definition for the angular2 ```$event``` object?


igor-ka:
@joehandlebar I might be wrong but I think that $event is an instance of an Event type(e.g., `($even: Event)`). It could have concrete type as well maybe like MouseEvent too.

I am not 100% sure about this, need to double check. Maybe Chrome Dev tools can show it at runtime.

app/src/components/dropdown/dropdown.ts
```diff
@@ -24,6 +26,6 @@ import {List} from 'immutable';
 })
 export default class Dropdown {
 
-  public owners: List<number>; 
+  public items: List<string>;
```
igor-ka: we can omit public keyword as it is public by default

joehandlebar:
updates on 1cda176

- removed ```public``` keywords
- when the owner name on a task is edited, the owner to filter on the view is updated too
- looks like ```$event``` has a type of ```Event``` but I didn't declare it b/c the ts definition for ```EventTarget``` doesn't have ```value```, which we're using atm

thanks @igor-ka

igor-ka:
@joehandlebar yes suk, public in the constructor should remain, as this syntax automatically create an instance variable (the clarification for that in the TS section), I can clarify this further if you want.

With regards to Event not having a some definitions, maybe you need to use a more concrete type that inherits from Event and that will have it, but i am not sure (i am ok with it being w/o types for now)

PR #75, opened 2016-01-29
Update 11-migrate.md

jacobharasimo:
type-o


e-schultz:
@jacobharasimo thanks for pointing this out. Just did a big re-structure of the handouts, but applied your fix to it. 

Thanks, :+1:

PR #76, opened 2016-01-29
Stop Atom from generating transpiler output inline

grough:
There was some talk about this in the #dev channel. Not sure it affects anyone else but it's been annoying me. Taken from https://github.com/rangle/angular-redux-starter/pull/23


e-schultz:
:+1:

PR #80, opened 2016-02-01
Refactor existing features to use Redux

joehandlebar:
The app is using redux now:
- ```task-edit```, ```task-add```, ```main```, and ```summary``` are now containers because there are unique routes allocated to them.
- ```tasks-service``` is not injected into components, it's instantiated in ```actions/tasks```
- the Angular2 ```Router``` is injected into components b/c it can only be initialized w/ a root component.
- form state remains encapsulated with ng2 form features

Weirdness:
- I'm still using observables for the course. ```Router``` only works with a root component, so to add/edit i pass a callback that handles the redirect into the action. Add/edit action creators are dispatched in ```.subscribe```. There is probably a cleaner way to do this (like have the action return a promise), but I don't know enough about observables.


e-schultz:
@joehandlebar can you rebase and resolve the conflicts on this - going to be doing a proper review of this PR today.

e-schultz:
Redux version is now in the branch [redux-refactor](https://github.com/rangle/ngCourse2/tree/redux-refactor), closing this PR.

PR #81, opened 2016-02-01
Require login to access Todo features

joehandlebar:
The app now has a simple login feature with generic username: ```user``` and password: ```pass```.

Redirection logic is handled in ```constructor```'s.

### DIFF
--- a/app/src/components/login/login.ts
+++ b/app/src/components/login/login.ts
@@ -1,0 +1,29 @@
+import {Component} from 'angular2/core';
+import {AuthService} from '../../services/auth-service';
+const TEMPLATE = require('./login.html');
+
+@Component({
+  selector: 'login',
+  template: TEMPLATE
+})
+export default class LoginComponent {
+  message: string;
+
+  constructor(public authService: AuthService) {
+    this.message = '';
+  }
+
+  login(username: string, password: string): boolean {
+    this.message = '';
+    if (!this.authService.login(username, password)) {
+      this.message = 'Incorrect credentials.';
+      return false;
+    }
+    return true;
+  }
+
+  logout(): boolean {
+    this.authService.logout();
+    return false;
+  }
+}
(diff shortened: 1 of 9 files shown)


app/src/components/login/login.ts
```diff
@@ -0,0 +1,49 @@
...
+
+  message: string;
+  loginForm: ControlGroup;
+
+  constructor(
+    public authService: AuthService,
+    private _builder: FormBuilder,
+    private _router: Router
+  ) {
+    this.message = '';
```
e-schultz: instead of having that here, on line 21, could have message:string = '';

app/src/services/auth-service.ts
```diff
@@ -0,0 +1,30 @@
...
+
+  logout(): any {
+    localStorage.removeItem('username');
+  }
+
+  getUser(): any {
+    return localStorage.getItem('username');
+  }
+
+  isLogged(): boolean {
```
e-schultz: I'd rename this to isLoggedIn to be a little more explicit - would need to update it in the other components/template also. Minor nitpick, but otherwise looking ok.

joehandlebar:
thanks @e-schultz. d19b099
- set login ```message = ''``` in class definition
- changed ```isLogged``` to ```isLoggedIn```

e-schultz:
Getting this error in the console when running
screen shot 2016-02-02 at 12 10 32 pm

e-schultz:
Looks like it happens if you try and navigate to http://localhost:3000/tasks without there being a username in local storage. Instead of bouncing to the login page correctly, things go sideways.

e-schultz:
@joehandlebar looks like doing a router.navigate in the constructor is causing the issue. If you put the logic into the ngOnInit lifecycle hook, seems to fix the issue.

```ts
export default class Main {

  constructor(
    public authService: AuthService,
    private _router: Router
  ) {

  }

  ngOnInit() {
    if (!this.authService.isLoggedIn()) {
      this._router.navigate(['/Login']);
    }

  }
}
```

joehandlebar:
@e-schultz  :+1:  thanks!

joehandlebar:
d19b099

- handle login redirects in ```ngOnInit```

e-schultz:
:shipit:

PR #87, opened 2016-02-03
Added testing components

tyllrb:
Here is the section on testing Angular 2 components.


igor-ka:
@tyllrb this is another excellent section.

It does get thick at times, I am wondering if we can "chew it out" a bit more to make it easier to understand and follow. (This is work that can be done in the future).

With regards to services tests at the bottom, I am wondering if we want to break it apart further? Testing components as well as sinon together gets quite heavy. Let me know what you think.

handout/testing/Unit Testing Components.md
```diff
@@ -0,0 +1,240 @@
...
+*quote.spec.ts*
+
+``` typescript
+class MockQuoteService {
+  public quote: "Test quote";
+  
+  getQuote() {
+  	return this.quote;
+  }
+}
```
barretodavid: Is not a problem that `MockQuoteService` is returning a plain value when the original service is returning a promise?

handout/testing/Unit Testing Components.md
```diff
@@ -0,0 +1,240 @@
...
+} from 'angular2/testing';
+
+describe("Testing Quote Component", () => {
+    beforeEachProviders(() => {
+		provide(QuoteService: {useClass: MockQuoteService})
+	});
+    
+it("Should get quote", injectAsync([TestComponentBuilder], (tcb) => {
+  return tcb.createAsync(QuoteComponent).then(fixture) => {
+  		fixture.detectChanges();
```
barretodavid: Do you know why is this first `detectChanges` needed? I understand the need of the second one that is trying to update the view with the new value of `this.quote`, but why the first one?

handout/testing/Unit Testing Components.md
```diff
@@ -0,0 +1,240 @@
...
+
+`TestComponentBuilder` will return a *fixture* for us to use in our tests. A fixture is powerful tool athat allows us to query the DOM rendered by a component as well as change DOM elements and component properties. It is the main point of testing components and we use it extensively. Here we have gotten access to our component through the `fixture.debugElement.componentInstance` property where we call `getQuote` to kickstart our only action in the `QuoteComponent` component. We call `fixture.detectChanges` to watch for any changes in the DOM. We then get access to the native underlying DOM elements through the `fixture.debugElement.nativeElement` property. Now we can check to see if the DOM rendered by our `QuoteComponent` contains the quote that we mocked in through the `QuoteService`. The final line, attempts to assert that the DOM's div tag contains the mocked quote 'Test Qoute' inside. If it does, then our component passes the test and works as expected, if it doesn't, that means our component is not outputting quotes correctly.
```
barretodavid: "We call `fixture.detectChanges` to watch for any changes in the DOM". Maybe we should avoid the word **watch** because it has a specific connotation in Angular 1. How about just saying "We call `fixture.detectChanges` to update the DOM with the new value of `this.quote`"?

handout/testing/Unit Testing Components.md
```diff
@@ -0,0 +1,240 @@
...
+  providers: [ExampleService]
+});
+class SimpleComponent() {}
+```
+
+This won't work when using `beforeEachProvider`, instead we can use the `TestComponentBuilder` to explicitly inject the `ExampleService` provider through `overrideProviders`. 
+
+``` typescript
+it('Should work', injectAsync([TestComponentBuilder], (tcb: TestComponentBuider) => {
+  tcb.overrideProviders(SimpleComponent, [ExampleService]).createAsync(SimpleComponent).then(fixture => {
```
barretodavid: Is `ExampleService` the real service or the mocked one?

handout/testing/Unit Testing Components.md
```diff
@@ -0,0 +1,240 @@
...
+  		// test all things relating to the message property here
+	});
+})
+```
+
+## Testing Asynchronous Actions
+
+Some components rely on asynchronous actions to work. This can be tricky to test, as we don't know exactly know when to look at the DOM for results. Fortuntely Angular provides a function `fakeAsync` which essential fakes asynchronous behaviour. FakeAsync will wrap our test in a zone, and then listen for any asynchronous operations, like setTimeouts, Promises, etc but will not actually call those functions asynchrounsly. Instead, it will rely on us calling the `tick` function which will call those functions immediatly and simulate time elapsing. 
+
+Suppose we had a component with a button, clicking that button will trigger a call to some service which will return data for the component to display. Heres what our unit test might look like:
```
barretodavid: It could be useful for the reader to show the implementation of this component

handout/testing/Unit Testing Components.md
igor-ka: :+1:

handout/testing/Unit Testing Components.md
```diff
@@ -0,0 +1,230 @@
...
+  	this.quoteService.getQuote().then((quote) => {
+  		this.quote = quote;
+	});
+   }
+}
+```
+
+This component relies on the `QuoteService` to get a random quote, which it will then display. The class is pretty simple, it only has the `getQuote` function that will modify the DOM, therefore it will be are main area of focus in testing.
+
+In order to test this component we need initiate the class `QuoteComponent`, the Angular testing library offers a way to inject any dependencies this class relies on, as well as a `TestComponentBuilder` object which will create the component for us and return a *component fixture* that we can perform tests on.
```
heykevingibson: "we need to initiate the class `QuoteComponent`. The Angular testing library offers a way to inject any dependencies...

handout/testing/Unit Testing Components.md
```diff
@@ -0,0 +1,230 @@
...
+	}
+  }));
+});
+```
+
+Testing the QuoteComponent is a fairly straightforward process. We want to create a QuoteComponent, feed it a quote and see if the DOM has changed such that the quote has shown up. This process requires us to create the component, pass in any dependencies, trigger the component to perform an action, and then look at the DOM to see if the action is what we expected. Lets take a look at how this is accomplished with the above unit test. 
+
+We use `beforeEachProviders` to feed in any dependencies that our component requires. Here our component depends on the `QuoteService` to get data, we mock this data ourselves as we can have control over what value we expect to show up. It is good practise to separate component testing from service testing - this makes it easier to test as you are only focusing on a single aspect of the application at a time. If your service fails, or your component fails, how will you know which one was the culprit? Using `provide` we inject the `QuoteService` dependency using our mock class `MockQuoteService` , where we will provide mock data for the component to consume.
```
heykevingibson: practice

handout/testing/Unit Testing Components.md
```diff
@@ -0,0 +1,230 @@
...
+
+`TestComponentBuilder` will return a *fixture* for us to use in our tests. A fixture is powerful tool that allows us to query the DOM rendered by a component as well as change DOM elements and component properties. It is the main point of testing components and we use it extensively. Here we have gotten access to our component through the `fixture.debugElement.componentInstance` property where we call `getQuote` to kickstart our only action in the `QuoteComponent` component. We call `fixture.detectChanges` to keep an eye out for any changes taking place to the DOM. We then get access to the native underlying DOM elements through the `fixture.debugElement.nativeElement` property. Now we can check to see if the DOM rendered by our `QuoteComponent` contains the quote that we mocked in through the `QuoteService`. The final line, attempts to assert that the DOM's div tag contains the mocked quote 'Test Quote' inside. If it does, then our component passes the test and works as expected, if it doesn't, that means our component is not outputting quotes correctly.
```
heykevingibson: A fixture is a powerful tool..

PR #96, opened 2016-02-17
Exercises - Routing - Basic, and Basic with Params

e-schultz:
@igor-ka @bennett000 @winkerVSbecks 
Initial two exercises for Routing, wondering if 1.0 and 1.1 is a little too granular - and if I should just make 1.1 become 1.0 that covers routing + params, then have a 2.0 that gets into child routes, and route lookup, then do a 3.0 that uses aux routes?


winkerVSbecks:
@e-schultz I agree, you could combine 1.0 and 1.1 but, this works too.

bennett000:
I don't think having granular exercises is necessarily a bad thing.  I could see it annoying some people, but they can just skip/skim.  I can also see it benefitting other people.

I also now realize that my exercise numbering is flawed

e-schultz:
Think I'll leave it as-is for now, the child routing one is taking me more time to setup than expected. Might come back later and change this one after the fact.

PR #97, opened 2016-02-18
Adds ES6 Exercises

bennett000:
I just looked at @e-schultz 's PR and I'm thinking I should redo my numbering scheme.  I also see the feature exercises as being optional or only for people that need them since they're so basic.

Tagging @winkerVSbecks for his opinion too.


bennett000:
I ended up correcting all the numbers since I needed to _force_ SystemJS to transpile the feature examples that do not use `import`/`export`.

Turned out that SystemJS was [_not_ compiling the examples](http://stackoverflow.com/questions/30720963/jspm-does-not-transpile-code-from-es6-to-es5 "System JS For The Win")

winkerVSbecks:
LGTM :shipit:

bennett000:
Great.  I'm going to round out the TS examples before I merge.

PR #103, opened 2016-02-19
DI exercises

tyllrb:
OK so here is a simple dependency injection exercise. Basically you have to create a root *Car* component that depends on three classes - *Tires*, *Body*, *Engine*. Each of these classes has 1 or 2 default properties, you need to inject these classes into the *Car* component and print out the value of each property.


bennett000:
@tyllrb I'm not sure how important this is, but your filenames have spaces.  Can you please make them follow the same pattern as the existing exercises?

Exercise: `1.0-some-title`
Solution: `1.0-some-title_solution`

PR #106, opened 2016-02-22
App unit testing

xiongemi:
- in package.json, added jasmine, updated angular and its dependency framework version number to remove warnings
- in karma.conf.js, changed the framework to jasmine
- updated tests.entry.ts to include the setup
- changed the simple.spec.ts to match the handout
- wrote unit tests for components, services, pipes
- add a folder mocks in root src directory to have mock providers


winkerVSbecks:
cc: @e-schultz @bennett000 @tyllrb

app/src/actions/counter.spec.ts
```diff
@@ -0,0 +1,64 @@
...
+import {INCREMENT_COUNTER, DECREMENT_COUNTER} from './counter';
+import * as CounterActions from './counter';
+import {
+  describe,
+  it
+} from 'angular2/testing';
+
+describe('counter action creators', () => {
+  it('increment should create INCREMENT_COUNTER action', () => {
+    chai.expect(CounterActions.increment())
```
e-schultz: If we are moving this from `Mocha` to `Jasmine`, should we still be using `Chai`?

app/src/components/task-edit/task-edit.spec.ts
```diff
@@ -0,0 +1,68 @@
...
+
+  beforeEachProviders(() => {
+    return [
+      mockTasksService.getProvider(),
+      FormBuilder,
+      mockRouterProvider.getProviders()
+    ];
+  });
+
+  it("Input should be populated with task", injectAsync(
```
e-schultz: Very nitpicky, but generally use `'` instead of `"`, 

```js
it('Input should be populated with task',
```

app/src/actions/counter.spec.ts
e-schultz: note: more of a question, if chai works with this still - I'm fine with that.

nexus-uw: from what I seen, 'angular2/testing' doesn't provide the usual jasmine 'assertions' (ie: expect(1).toBe(1)). including chai.expect related stuff works great in angular 2 (plus it was what we used for angular 1).


The big thing to remember is that we can't have both Jasmine and Mocha on the page at the same time.

Issue #105, opened 2016-02-22
tsd@0.6.5 has been deprecated


nexus-uw:
running ```npm install```
results in 
```bash
npm WARN deprecated tsd@0.6.5: TSD is deprecated in favor of Typings (https://github.com/typings/typings) - see https://github.com/DefinitelyTyped/tsd/issues/269 for more information
```

e-schultz:
migrated to typings in master, and in the redux version of the app

PR #116, opened 2016-02-25
[#105] Update deps and move from TSD to Typings

e-schultz:
@bennett000 @winkerVSbecks moved from TSD to Typings, using cross-env for setting env settings (should help with windows issues), and updated to beta 7


bennett000:
This is great, I was actually _just_ thinking about this and how I need to update the ng-upgrade example to use typings.

PR #119, opened 2016-02-25
Chore refactor task crud component

e-schultz:
@bennett000 @winkerVSbecks @igor-ka Didn't really like how the task-add and task-edit containers were so similar in terms of template / code, and didn't really demonstrate smart/dumb very well.

Created a task-form component, and then refactored task-add and task-edit to use it.


bennett000:
LGTM :shipit:

PR #120, opened 2016-02-26
WIP - Redux workshop 1

e-schultz:
@bennett000 @winkerVSbecks 

Starting to build out the 'kitchen sink' part that should show most if not all of the dumb components. Think a few of them will need a bit of refactoring to be able to get them all here, and/or move out some of the 'smarter' ones (ie: login) to being a container. 

The idea is workshop-0 - would have  app skeleton, a sample component, a little bit of kitchen sink filled out - then go through building out the dumb components. Then workshop-1 would pick up from here, components all done - then start building out the other parts of the app.


e-schultz:
closing PR, made some orphan branches for workshop stuff.

PR #122, opened 2016-02-29
Update ES6 ...rest exercise

e-schultz:
Adjusted the problem and solution to better demonstrate `...rest` for
arguments.

@bennett000 just updated it a bit to better show how multiple arguments get used with `...`


bennett000:
:shipit:

PR #123, opened 2016-02-29
Refactoring Login Component

e-schultz:
* Create form that just renders the login form
* Create container to handle login control flow

@winkerVSbecks or @bennett000 for review

### DIFF
--- a/app/src/components/login-form/login-form.ts
+++ b/app/src/components/login-form/login-form.ts
@@ -1,0 +1,45 @@
+import {Component, Input, Output, EventEmitter} from 'angular2/core';
+import {Router} from 'angular2/router';
+import {
+  FORM_BINDINGS,
+  FORM_DIRECTIVES,
+  ControlGroup,
+  FormBuilder,
+  Validators
+} from 'angular2/common';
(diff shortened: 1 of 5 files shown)


app/src/components/login-form/login-form.ts
```diff
@@ -0,0 +1,45 @@
...
+
+@Component({
+  selector: 'login-form',
+  directives: [FORM_DIRECTIVES],
+  viewBindings: [FORM_BINDINGS],
+  template: TEMPLATE
+})
+export default class LoginForm {
+
+  @Input() message: string = '';
```
winkerVSbecks: @e-schultz should we switch to `inputs` and `outputs` everywhere?

winkerVSbecks:
Just that one question about `@Input` vs `inputs` Otherwise :shipit:

app/src/components/login-form/login-form.ts
e-schultz: thinking we should switch to `@Input` and `@Output` to be consistent - going to be a bit of an ongoing refactor with that. Going to try and keep those commits cherry-pickable so I can apply them to a few branches easily.

PR #124, opened 2016-02-29
Chore refactor task filters

e-schultz:
@bennett000 @winkerVSbecks 
Was playing around with the `kitchen-sink` in another repo, and noticed that when passing callbacks as inputs, the context of 'this' gets weird. 

when in the callback in kitchen-sink, 'this' was pointing to TaskFilter, even when trying to change the callback to be defined like `myCb = (x:any) => { }`

Changing things to event emitter started to get the expected behavior.


winkerVSbecks:
Makes sense. We should probably be using event emitter instead of callbacks anyway. I think it makes for cleaner API.

:shipit:

PR #126, opened 2016-03-01
Redux simple

e-schultz:


e-schultz:
@bennett000 @winkerVSbecks kept running into annoyances while trying to setup the 'workshop' version of the app. WIP on some cleanup/refactoring 

Trying to keep stuff under components pretty dumb - including not hard-coding routes/etc that it might need and passing them in.

Also trying to change things to be 'pass data down, emit events up' and being more consistent with `@Input` and `@Output`

Also renamed a few things to make a little more sense, ie: 'Main' got renamed to 'Tasks', since it is the task home/etc.

Tempted for the 'workshop' version of the app - leaving out the login flow, and possibly leaving out the API and just working in-memory to keep the action creators simple. Then, maybe as a final step if there is time - work into using the API. 

Also thinking of removing most(?) of the pipes - and using the mapStateToDispatch to do things like filtering out tasks/etc, and to hopefully better demonstrate 'deriving new data to support new features' from a global app state.

PR #127, opened 2016-03-01
Refactor task actions to not take callback

e-schultz:
Previous version had you passing in a callback for the success of an API call. I didn't like this, and wanted to switch it to returning the observable sequence.

Initially I tried just doing

```
x = task.add(x)
x.subscribe()
return x;
// in my component
x.subscribe()
```

but it would fire off the HTTP request twice. This seems to be giving me the results I expect, but is this the right way to be doing it?

@AbdellaToronto @igor-ka @winkerVSbecks @bennett000 

Still a little rusty with RxJs Observables - feedback?

### DIFF
--- a/app/src/actions/tasks.ts
+++ b/app/src/actions/tasks.ts
@@ -1,9 +1,14 @@
-import TasksService, {Task, TaskMap} from '../services/tasks-service'; 
-import * as FilterActions from './filters';
+import TasksService, {Task, TaskMap} from '../services/tasks-service';
+import {selectOwner}  from './filters';
 import {Http} from 'angular2/http';
 import {Injector, provide} from 'angular2/core';
 import {HTTP_PROVIDERS} from 'angular2/http';
 import {Map} from 'immutable';
+import 'rxjs/add/operator/filter';
+import 'rxjs/add/operator/pluck';
+import 'rxjs/add/operator/map';
+import 'rxjs/add/operator/publish';
+import 'rxjs/add/operator/share';
 
 const injector = Injector.resolveAndCreate([
   HTTP_PROVIDERS,
@@ -19,8 +24,8 @@ export const TASK_UPDATED = 'TASK_UPDATED';
 export const TASK_MARKED = 'TASK_MARKED';
 
 export function loadTasks() {
-  return (dispatch) => Tasks.fetch()   
-    .subscribe((res: Array<Task>) => { 
+  return (dispatch) => Tasks.fetch()
+    .subscribe((res: Array<Task>) => {
       dispatch({
         type: TASKS_LOADED,
         payload: res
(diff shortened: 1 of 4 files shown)


app/src/actions/tasks.ts
```diff
@@ -31,15 +33,24 @@ export function loadTasks() {
...
+          payload: res[0]
+        });
+      }
+    })
+
+
+    return taskAdd$;
+
+
+  }
```
AbdellaToronto: So I'm thinking about this solution, and it seems like it should work fine - but there may be a more elegant way about it - if I am not making any mistakes about the relationship between actions and reducers. It would also mean 

First idea - probably the least amount of change and entirely superficial:

```javascript

taskAdd$
.filter(res => res.length === 1)
.pluck('0') //Probably gratuitous
.subscribe(res => dispatch({type: TASK_ADDED, payload: res}); 
return taskAdd$;
```

You can see, not super essential, but maybe looks a bit nicer and is a little more functionally declarative.

The more interesting change (in my opinion) would be having the reducer subscribe to the task observables - but the more I think about it, the more I realize just how complex that would be. Still interesting, but I think something that could potentially be explored with the ng2ReduRx re-write.

app/src/actions/tasks.ts
```diff
@@ -31,15 +33,24 @@ export function loadTasks() {
...
-      .subscribe((res) => {
-        if (res.length === 1) {
-          dispatch({
-            type: TASK_ADDED,
-            payload: res[0]
-          });
-        }
-      }, null, onComplete); }
+
+    const taskAdd$ = Tasks.add(task).publish();
```
AbdellaToronto: I would probably push the publish up into the service - and unless there is a case I am not considering, you can get rid of connect() if, instead of publish, you just use .share() - it's basically a publish/connect (well not connect, refcount, but close enough)

app/src/actions/tasks.ts
e-schultz: > The more interesting change (in my opinion) would be having the reducer subscribe to the task observables 

With Redux - the state/actions are based into the reducers by an outside thing. WIth more traditional flux, you do have a 'store listens to x' and sort of reaches outside of itself. With Redux - things are passed in. So having your reducer explicitly be subscribed to a stream is sort of going against that.

app/src/actions/tasks.ts
e-schultz: I'll give that a try later today and let you know how it goes, didn't know about .share vs .publish, but pushing it up into taskservice does seem to be cleaner - and not needing to worry about it in other parts of the code that might be consuming it.

I'll chat over lunch - but thanks.

app/src/actions/tasks.ts
```diff
@@ -31,15 +33,24 @@ export function loadTasks() {
...
+    taskAdd$.connect();
```
winkerVSbecks: Not sure this is necessary. The service returns an observable. You can subscribe to it right away or attach operators like Abdella suggested and then subscribe.

e-schultz: the issue I was having - was once the task was added, wanted to .subscribe in the component to redirect to the correct page. Without using .publish and/or .share - the second .subscribe would trigger another request. 

Talked with Della over lunch, and have an idea of how to handle this now

e-schultz:
@AbdellaToronto @winkerVSbecks cleaned up based on convo I had with della

winkerVSbecks:
:shipit:

PR #128, opened 2016-03-02
Added a for(let x...) example

e-schultz:


e-schultz:
@bennett000 small add to ES6 to show `let` in a for loop

PR #129, opened 2016-03-02
Add example to show destructuring a func arg

e-schultz:
@bennett000 @winkerVSbecks added quick example to show how to use destructuring for a function argument


winkerVSbecks:
:shipit:

PR #131, opened 2016-03-03
Fix first few typos in Typescript section

JaKXz:
can @e-schultz review?

### DIFF
--- a/handout/features/working_with_tsc.md
+++ b/handout/features/working_with_tsc.md
@@ -1,34 +1,34 @@
 # Working With `tsc`
 
 So far `tsc` has been used to compile a single file.  Typically developers have
-a _lot_ more than one file to compile.  `tsc` can handle multiple files as
+a _lot_ more than one file to compile. Thankfully  `tsc` can handle multiple files as
 arguments.
 
 Imagine two ultra simple files/modules:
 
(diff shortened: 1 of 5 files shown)


handout/features/working_with_tsc.md
```diff
@@ -1,34 +1,34 @@
...
 So far `tsc` has been used to compile a single file.  Typically developers have
-a _lot_ more than one file to compile.  `tsc` can handle multiple files as
+a _lot_ more than one file to compile. Thankfully  `tsc` can handle multiple files as
 arguments.
 
 Imagine two ultra simple files/modules:
 
 a.ts
-```js
+```ts
```
e-schultz: the highlighting with git-book gets finicky when you use ts instead of js.

JaKXz: > Specify the v2 of plugin-highlight in your book.json and check if it helps.

From git book support, let's try that?

JaKXz: Verified locally, thanks to [these updates to the highlight plugin](https://github.com/GitbookIO/plugin-highlight/compare/1.0.3...2.0.0) that Gitbook support made when I emailed them. Props to those guys.

PR #150, opened 2016-03-16
Incorporates testing of components, services, and redux into the book

bennett000:


bennett000:
Just reviewed with Evan in person, self merging...

PR #157, opened 2016-04-04
Change default port for gitbook to 3000

gusgard:
When start the server gitbook is listening in the port 4000 and in the README it says 3000.


e-schultz:
the demo app in the app folder runs off of port 3000, I'll update the readme.md to make it more clear on running the book locally vs running the app.

Thanks for the pr though, thanks.

Evan

PR #171, opened 2016-04-17
Fix for type errors in TS exercise 1 solution.

### LINKED ISSUE #107: features/typescript - first exercise is incorrect
The first exercise in the features/typescript exercise section does not work - there is no correct solution.

SethDavenport:
Fixes #107


SethDavenport:
@bennett000 @tyllrb

PR #224, opened 2016-05-12
update exercises to rc1 (except redux)

mdegani:
Current issues (I'm working on them):
- getting an HTTP error when the box is cleared in observable exercise -- this problem existed before
  these changes
- routing back button doesn't work -- problem is new, back button
  worked before these changes


bennett000:
:shipit: but please file an issue regarding the router problem. We will need to redo the section when the new documentation comes out.

PR #226, opened 2016-05-12
fix function interface example

mdegani:
please see http://jsbin.com/tupaqikehu/edit?js,console for working example.


bennett000:
:shipit:

Issue #221, opened 2016-05-11
update ngFors in handouts


mdegani:
new syntax for ngFor: `<div *ngFor="let hero of heroes">{{ hero }}</div>`

jamesbs:
This has been addressed in #229 can you review @bennett000?

bennett000:
This has been merged.  It missed a connection so it did not auto close.  Thanks for the PR.

PR #228, opened 2016-05-12
update ngFor's in the handouts

mdegani:
Reference: https://github.com/angular/angular/commit/27a7b51

connected to https://github.com/rangle/ngCourse2/issues/221


bennett000:
This all makes sense.  @jamesbs where do you stand on this since you're updating the example code?

PR #232, opened 2016-05-17
Adds a shoutout to ngrx

bennett000:


bennett000:
:shipit: @AbdellaToronto was able to take a quick look ad hoc.  Quasi-self shipping this one.

PR #239, opened 2016-05-18
Hamburger must be of type string in this case

un33k:
Hamburger must be of type string in this case


bennett000:
@un33k can you point me to documentation or source code that verifies that Hamburger must be a string in this case?  If that's the official documentation then we'll change it.

For the moment though it works fine using the class symbol (assuming I've reproduced the plunker correctly): https://plnkr.co/edit/jYBKc4V2KP8Z0oKOgjsb?p=preview

un33k:
In the docs Hamberger is not imported. In Plunker it is.

bennett000:
We're going to fix this with #247 and keep the class reference.  TO un33k's point we've imported `Hamburger`.  Thanks!

PR #246, opened 2016-05-26
adjust intro: book is course material and stand-alone

mdegani:


bennett000:
Turns out this wasn't the landing page we needed to edit, but these edits LGTM, :shipit:

PR #253, opened 2016-06-07
Fix book iframes

### LINKED ISSUE #252: Omit iframes from ebook, and add feature to optionally omit from pdf
Since iframes don't render in ebook formats, and only some websites work in the pdf.

steven-ks:
Fixes #252 

@winkerVSbecks

### DIFF
--- a/book.json
+++ b/book.json
@@ -4,6 +4,10 @@
     "theme-default",
     "ga"
   ],
+  "styles": {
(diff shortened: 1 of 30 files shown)


winkerVSbecks:
:shipit:

steven-ks:
@winkerVSbecks Should I have access to merge PRs, or will someone else do it? Right now I don't have that capability.

PR #270, opened 2016-06-14
Update basic router example to router 3

e-schultz:
@winkerVSbecks @bennett000 @SethDavenport 

Playing with new router and moving an example from deprecated to router 3. 

Main changes:

* What was once in `@RouteConfig` is now just an array
* That gets fed into the `provideRouter` during bootstrap 
* routerLink DSL is mostly the same, except the `as` from the config is gone - and seems like you need to give it a path, instead of a name that it's registered as.


andrewlo:
@e-schultz This should probably be closed since I added all the new examples for the v3 router

e-schultz:
👍 yup

Issue #280, opened 2016-06-17
errors on the build process


MichaelLSmith:
after running npm install in the app directory, I tried running npm run dev. I get all these errors after the last successful line 497. Am I missing something obvious?

 [497] ./~/redux-logger/lib/index.js 8.49 kB {1} [built] -- worked
ERROR in /home/michael/git/rangle_angular2/ngCourse2/app/src/mocks/mock-router-provider.ts
(41,7): error TS2346: Supplied parameters do not match any signature of call target.

ERROR in ./src/components/task-edit/task-edit.ts
(8,3): error TS2305: Module '"/home/michael/git/rangle_angular2/ngCourse2/app/node_modules/@angular/common/index"' has no exported member 'FORM_BINDINGS'.

ERROR in ./src/components/task-edit/task-edit.ts
(20,3): error TS2345: Argument of type '{ selector: string; directives: (typeof RouterLink | any[])[]; viewBindings: any[]; template: any...' is not assignable to parameter of type '{ selector?: string; inputs?: string[]; outputs?: string[]; properties?: string[]; events?: strin...'.
  Object literal may only specify known properties, and 'viewBindings' does not exist in type '{ selector?: string; inputs?: string[]; outputs?: string[]; properties?: string[]; events?: strin...'.

ERROR in ./src/components/task-add/task-add.ts
(4,3): error TS2305: Module '"/home/michael/git/rangle_angular2/ngCourse2/app/node_modules/@angular/common/index"' has no exported member 'FORM_BINDINGS'.

ERROR in ./src/components/task-add/task-add.ts
(16,3): error TS2345: Argument of type '{ selector: string; directives: (typeof RouterLink | Type[])[]; viewBindings: any[]; template: an...' is not assignable to parameter of type '{ selector?: string; inputs?: string[]; outputs?: string[]; properties?: string[]; events?: strin...'.
  Object literal may only specify known properties, and 'viewBindings' does not exist in type '{ selector?: string; inputs?: string[]; outputs?: string[]; properties?: string[]; events?: strin...'.

ERROR in ./src/components/login/login.ts
(4,3): error TS2305: Module '"/home/michael/git/rangle_angular2/ngCourse2/app/node_modules/@angular/common/index"' has no exported member 'FORM_BINDINGS'.
webpack: bundle is now VALID.

e-schultz:
@MichaelLSmith thanks for pointing this out, there were some changes in RC2 that broke things, and we were not locking down the version to RC1. 

[...]

PR #284, opened 2016-06-18
fix rc2 breaks - #280

e-schultz:
* FORM_BINDINGS -> FORM_PROVIDERS
* viewBindings: -> viewProviders

These will eventually be deprecated,

> It looks like you're using the old forms module. This will be opt-in in
> the next RC, and
>      will eventually be removed in favor of the new forms module. For
> more information, see:
>      https://docs.google.com/document/u/1/d/1RIezQqE4aEhBRmArIAS1mRIZtWFf6JxN_7B4meyWK0Y/pub

Will need to update again later, but this will at least let the app run
/ tests run and pass now.

### DIFF
--- a/app/src/components/login/login.ts
+++ b/app/src/components/login/login.ts
@@ -1,7 +1,7 @@
 import {Component} from '@angular/core';
 import {Router} from '@angular/router-deprecated';
 import {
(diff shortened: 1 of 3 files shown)


SethDavenport:
:shipit:

PR #291, opened 2016-06-20
LL-Update refresher_on_this.md

llittlewtw:
copy edits
I set up the two cases as a numbered list, please confirm that I've set up the list and indent spacing properly. I'm not sure if the info following #2's example should actually be indented as part of #2.


SethDavenport:
The 'view example' should not be indented, since it demonstrates to both points.

SethDavenport:
Otherwise looks good.

PR #292, opened 2016-06-20
LL-Update arrow_functions.md

llittlewtw:
copy edits

### DIFF
--- a/handout/features/arrow_functions.md
+++ b/handout/features/arrow_functions.md
@@ -1,8 +1,7 @@
 # Arrow Functions
 
 ES6 offers some new syntax for dealing with `this`: "arrow functions".  
-Arrow functions also make working with "higher order" functions (functions that take
-functions as parameters) much easier to work with.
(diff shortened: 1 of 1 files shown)


handout/features/arrow_functions.md
```diff
@@ -1,7 +1,7 @@
 # Arrow Functions
 
-ES6 offers some new syntax for dealing with `this`: "arrow functions".  Arrow
-function also make working with "higher order" functions (functions that take
+ES6 offers some new syntax for dealing with `this`: "arrow functions".  
+Arrow functions also make working with "higher order" functions (functions that take
```
SethDavenport: '.. make working with ... much easier to work with'

Probably a place where our text could be refined even more :)

llittlewtw: safe to just remove the parenthetical?

SethDavenport: What I'm getting at is the two instances of 'work with'

Maybe something like: "Arrow functions also make higher order functions much easier to work with."

PR #294, opened 2016-06-20
LL-Update constants,_and_block_scoped_variables.md

llittlewtw:
copy edits

### DIFF
--- a/handout/features/constants,_and_block_scoped_variables.md
+++ b/handout/features/constants,_and_block_scoped_variables.md
@@ -60,7 +60,7 @@ for(let x=0; x<5; x++) {
 ```
 
 Alternatively, `const` is read-only.  Once `const` has been assigned, the identifier cannot be
-reassigned; the value is [not immutable](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/const).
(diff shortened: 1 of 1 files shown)


handout/features/constants,_and_block_scoped_variables.md
```diff
@@ -41,29 +41,26 @@ console.log(j); // 9
...
   setTimeout(()=>console.log(x), 0)
 }
 ```
 
-
-Alternatively, `const` is read only.  Once `const` has been assigned, the identifier can not be
-re-assigned, the value is [not immutable](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/const).
-
+Alternatively, `const` is read-only.  Once `const` has been assigned, the identifier cannot be
+reassigned; the value is [not immutable](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/const).
```
SethDavenport: Might be a bit clearer to write:

"... the identifier cannot be reassigned; however the value itself is still [mutable](..."

What we're getting at here is that some people mistakenly assume `const` results in value immutability; however in reality it only results in reference immutability.

SethDavenport:
:shipit:

PR #297, opened 2016-06-20
LL-Update destructuring.md

llittlewtw:
copy edits
** please check the last part of the last sentence ("dynamic destructing during for-ins iterators) - it seems greek to me, is it correct? Should it at least be changed to "destructuring"?

### DIFF
--- a/handout/features/destructuring.md
+++ b/handout/features/destructuring.md
@@ -54,6 +54,6 @@ sayName(john) // -> Helo John Smith Doe
 
 There are many more sophisticated things that can be done with destructuring,
 and the [mdn][mdnDest] has some great examples, including nested object
-destructuring and dynamic destructing during for-ins iterators.
+destructuring and dynamic destructuring with  `for ... in`  operators"..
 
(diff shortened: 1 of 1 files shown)


handout/features/destructuring.md
```diff
@@ -52,8 +52,8 @@ sayName(jane) // -> Hello Jane N/A Doe
 sayName(john) // -> Helo John Smith Doe
 ```
 
-There are _many_ more sophisticated things that can be done with destructuring,
-and the [mdn][mdnDest] has some great examples, including nested Object
-destructuring, and dynamic destructing during for-ins iterators.
+There are many more sophisticated things that can be done with destructuring,
+and the [mdn][mdnDest] has some great examples, including nested object
+destructuring and dynamic destructing during for-ins iterators.
```
SethDavenport: Yeah this does make sense I think to a JS programmer.

Perhaps:

"... dynamic destructuring with `for ... in` operators".

SethDavenport:
:shipit:

PR #302, opened 2016-06-20
adding in cover images that get applied to PDF, ePub and .mobi

driches:
Gitbooks automatically adds cover.jpg or cover_small.jpg to the exported file for PDF, ePUB and .Mobi. Adding these files in the correct dimensions.

http://toolchain.gitbook.com/ebook.html


SethDavenport:
:shipit:

PR #325, opened 2016-06-21
LL-Update file_structure.md

llittlewtw:
copy edits
please review markdown, I'm not sure why the last few paragraphs are showing as blue in the editor, I rearranged the examples but don't think I missed any pieces. It does seem to look ok in the preview.

### DIFF
--- a/handout/bootstrapping/file_structure.md
+++ b/handout/bootstrapping/file_structure.md
@@ -1,22 +1,10 @@
-# File structure
+# File Structure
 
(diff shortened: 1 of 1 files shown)


SethDavenport:
@llittlewtw which editor are you using?

llittlewtw:
I'm editing directly in Github, so whatever is the native editor. If it looks ok to you and looks ok in the output, we can ignore.

PR #330, opened 2016-06-21
LL-Update app_structure_with_components.md

llittlewtw:
copy edits

### DIFF
--- a/handout/components/app_structure_with_components.md
+++ b/handout/components/app_structure_with_components.md
@@ -16,5 +16,5 @@ For example consider the following:
 </TodoApp>
 ```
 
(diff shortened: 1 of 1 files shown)


handout/components/app_structure_with_components.md
```diff
@@ -15,5 +16,5 @@ For example consider the following:
 </TodoApp>
 ```
 
-At the root we have `TodoApp` which consists of a `TodoList` and a `TodoForm`. Within the list we have several `TodoItem`s. Each of these components are visible to the user and they can interact with these components and perform actions.
+At the root we have `TodoApp` which consists of a `TodoList` and a `TodoForm`. Within the list we have several `TodoItem`s. Each of these components is visible to the programmer, who can interact with these components and perform actions.
```
SethDavenport: In this case, we mean 'user', not 'programmer' - as in the end-user, or the person interacting with the finished software.

PR #345, opened 2016-06-21
LL-Update using_observables_from_other_sources.md

llittlewtw:
copy edits

### DIFF
--- a/handout/observables/using_observables_from_other_sources.md
+++ b/handout/observables/using_observables_from_other_sources.md
@@ -33,7 +33,7 @@ export class App {
 ```
 [View Example](http://plnkr.co/edit/FjJz4lTX2hhHkMu87cln?p=preview)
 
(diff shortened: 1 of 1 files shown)


handout/observables/using_observables_from_other_sources.md
```diff
@@ -33,11 +33,11 @@ export class App {
 ```
 [View Example](http://plnkr.co/edit/FjJz4lTX2hhHkMu87cln?p=preview)
 
-This basic examples outlines how the `Http` libraries common routines like `get`, `post`, `put`, and `delete` all return Observables that allow us to asynchronously process any resulting data. 
+This basic examples outlines how the `Http` libraries' common routines like `get`, `post`, `put`, and `delete` all return observables that allow us to asynchronously process any resulting data.
```
SethDavenport: I think libraries should be singular here.  As in "... the `Http` library's common routines ..."

PR #347, opened 2016-06-21
LL-Update combining_streams_with_flatmap.md

llittlewtw:
copy edits

### DIFF
--- a/handout/observables/combining_streams_with_flatmap.md
+++ b/handout/observables/combining_streams_with_flatmap.md
@@ -36,7 +36,7 @@ export class SearchService {
 
 Here we have a basic service that will undergo a search query to Spotify by performing a get request with a supplied search term. This `search` function returns an observable that has had some basic post-processing done (turning the response into a JSON object). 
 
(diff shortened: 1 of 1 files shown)


handout/observables/combining_streams_with_flatmap.md
```diff
@@ -34,9 +34,9 @@ export class SearchService {
 }
 ```
 
-Here we have a basic service that will undergo a search query to Spotify by performing a get request with a supplied search term. This `search` function returns an Observable that has had some basic post-processing done (turning the response into a JSON object). 
+Here we have a basic service that will undergo a search query to Spotify by performing a get request with a supplied search term. This `search` function returns an observable that has had some basic post-processing done (turning the response into a JSON object). 
 
-OK, lets take a look at the Component that will be using this service. 
+OK, lets take a look at the component that will be using this service.
```
SethDavenport: let's

llittlewtw: oops, thanks, missed that one!

PR #482, opened 2016-06-26
LL-Update class_decorators.md

llittlewtw:
changed Cdn spelling of behaviour in code example - please confirm that this is ok


SethDavenport:
Yes - lines that start with `//` are comments so this doesn't affect the execution of the code sample.

PR #543, opened 2016-06-28
LL-Update summary.md

llittlewtw:
remove serial comma (US spelling)


driches:
`They provide a multitude of functions to create stream from many other types, and to manipulate and transform them.`

Switches between stream singular and "them" referring to plural. 

Should read as follows:
They provide a multitude of functions to create streams from many other types, and to manipulate and transform them.

PR #563, opened 2016-06-28
LL-Update using-chai.md

llittlewtw:
remove serial comma (US spelling)


SethDavenport:
This PR makes sense in the context of copy editing, so I'll merge it. However it raises a broader question for the support team: does it even make sense to cover chai at all if we're using Jasmine?

Jasmine has its own expectation syntax. I typically consider Jasmine to be functionally a replacement for mocha+chai+sinon

SethDavenport:
@bennett000 @clbond @winkervsbecks

PR #585, opened 2016-06-29
LL-Update total_conversion.md

llittlewtw:
change title to gerund form


SethDavenport:
Not convinced that we want to title it 'avoiding'... that meaning is a little strong. This is more a note about it.

SethDavenport:
@bennett000 thoughts?

bennett000:
I also think it's too strong.  I'm going to close this one _without_ merging.

PR #596, opened 2016-07-07
Fixes error in SUMMARY.md and removes/skips intro image caption

bennett000:


bennett000:
self :shipit: for two reasons:

- the current state of the book is technically broken and it's not building because `SUMMARY.md` is pointing to a missing file
- the only way to confirm the landing page fix works is to update the book and find out.

SethDavenport:
_post-hoc_ :shipit: 😉

PR #603, opened 2016-07-08
Increase mentions of ngrx in Redux section

### LINKED ISSUE #597: Update Chapter 14 Redux to Redux + ngRx

jamesbs:
Fixes #597 
- Changed section name to Redux and Ngrx
- Added more content to README.md regarding ngrx
- Changed implementation agnostic sections to avoid mentioning Redux if it makes sense to do so
- Added a Redux/Ngrx tools section to summarize some of the things that are possible with ngrx without going into code examples

### DIFF
--- a/handout/redux/README.md
+++ b/handout/redux/README.md
@@ -1,4 +1,4 @@
-# Redux
+# Redux and Ngrx
 
 ## What is Redux?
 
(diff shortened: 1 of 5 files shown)


handout/redux/review_of_reducers_and_pure_functions.md
```diff
@@ -1,8 +1,8 @@
 # Review of Reducers and Pure Functions
```
yuri: Not a reason to block this PR, but this section seems to call for at least of tiny bit of extension. (Maybe we should just file this as an issue?)

yuri:
:shipit: 

This PR looks good to me and accomplishes its job, but it seems to me that the section that aims to explain the idea of a reducer ends too abruptly. I think even just adding a paragraph after the summation example would help. But this should be a separate issue and a separate PR.

handout/redux/README.md
```diff
@@ -10,12 +10,14 @@ Where Flux applications traditionally have multiple stores, Redux applications
...
+One popular Angular 2 specific implementation of the Redux pattern is [Ng2-Redux](https://github.com/wbuchwalter/ng2-redux), which we'll be using to describe how to use this approach in an application.
 
-In this course we discuss a more traditional Redux pattern using the
-[Ng2-Redux](https://github.com/wbuchwalter/ng2-redux) bindings library.
+
+## What is Ngrx?
+
+Redux implementation has been very well received and has inspired the creation of [ngrx](https://github.com/ngrx "ngrx collection"), a set of modules that implement the same way of managing state as well as some of the middleware and tools in the Redux ecosystem. Ngrx was created to be used specifically with Angular 2 and [RxJS](https://github.com/Reactive-Extensions/RxJS), as it leans heavily on the observable paradigm. 
+
+Although we'll be using Ng2-Redux, a lot of the same lessons apply with regards to ngrx though the syntax may be different and have some slight differences in what abstractions are involved.
```
e-schultz: Although we will be using ng2-redux, a lot of the same concepts apply in regards to reducers and action creators between redux and ngrx, though the syntax may be different and have some slight differences in what abstractions are involved.

e-schultz:
minor suggestion in wording, but overall it's good - and mentioning Ngrx is good, as it is a question that is coming up increasingly.

Issue #602, opened 2016-07-08
Upgrade to RC4 (or newer)


bennett000:
The following sections will need to be upgraded to RC4 (or whatever RC is the most stable/recent when the card gets implemented) 

- [ ] upgrade exercises
- [ ] upgrade examples
- [ ] upgrade app
- [ ] upgrade any/all examples in the book to reflect new changes in the RC

PR #605, opened 2016-07-11
increase versions of angular to rc4

mdegani:
connect to #602 

- did not change exercises in the forms or routing folders
- modified both starting point and solution.  All solutions worked without modification.


e-schultz:
👍

PR #607, opened 2016-07-12
update plunkers and examples

mdegani:
connect to #602 

- bumped version to rc4 on examples and made sure they still worked as expected.
- updated gitbook with new urls
- did not update: form builder, routing or redux examples


e-schultz:
:shipit:

PR #608, opened 2016-07-12
New router

andrewlo:


andrewlo:
This PR includes the rename of the old router, which has it's own PR here https://github.com/rangle/ngCourse2/pull/606

When reviewing please look at the top commit which adds the new chapter.

handout/routing/config.md
```diff
@@ -0,0 +1,63 @@
...
+
+RouteConfig is an array type defined in Angular 2 which takes the array of routes definitions to define the routes within the application. Each route could have different attributes; some of the common attributes are:
+
+* _path_ - URL to be shown in the browser when application is on the specific route
+* _component_ - Component to be rendered when the application is on the specific route
+* _redirectTo_ - Redirect route if needed; each route can have either component or redirect attribute defined in the route (covered later in this chapter)
+* _pathMatch_ - Optional property that defaults to 'prefix'; determines whether to match full URLs or just the beginning. Set to 'full' when defining the redirect route.
+* _children_ - Array of route definitions objects representing the child routes of this route (covered later in this chapter)
+
+## provideRouter ##
```
e-schultz: Wondering if this could be simplified for now to something like

```js
 const routes: RouterConfig = [
   { path: 'component-one', component: ComponentOne },
   { path: 'component-two', component: ComponentTwo }];

bootstrap(AppComponent,[provideRouter(routes)]); 
```

It's a little more explicit and can grok in one pass instead of needing to jump between a few different snippits to see what is going on.

When we expand on the section, could have a 'best practices around organizing routes' that starts to use the pattern that you initially had here.

handout/routing/redirects.md
```diff
@@ -0,0 +1,16 @@
+# Route Redirections #
```
e-schultz: I think we should cover `pathMatch: 'full'` and `pathMatch: 'prefix'` - I can see this being a question that most people would ask after reading this section if we didn't touch on it.

handout/routing/child_routes.md
```diff
@@ -1,60 +1,92 @@
...
+    ]
+  }
+];
 ```
-[View Example](https://plnkr.co/edit/qCYF138UXe6l9CiTXggG?p=preview)
+Since the `Overview` child route of `product-details` has an empty path, it will be loaded by default. The `specs` child route remains the same.
+
+[View Example with child routes](https://plnkr.co/edit/QlMe6pMINxJGTdA3xm0B?p=preview)
+
+[View Example with route params & child routes](https://plnkr.co/edit/g88pvg4LQq7XVJo7iG9b?p=preview)
```
e-schultz: Was playing around with an example to show the difference between

```html
<a [routerLink]="route"></a>
<a [routerLink]="/route"></a>
<a [routerLink]="../route"></a>
```

Might be worth trying to incorporate the lookup of 'current vs parent vs root' routes

e-schultz:
@andrewlo 👍 - could you do another PR where you bring in your plunkers into the examples folder - plunker has a 'download as zip' which is how we got them in before, so should be easy. 

Next step is to work out some exercises, which can also be in a separate PR.

PR #611, opened 2016-07-14
Added a Getting Started, a new blurb for the Form page, a Template-Dr…

seanmay:
…iven Forms, a FormBuilder page, and updated SUMMARY


e-schultz:
@seanmay good start, think I'll wait until the form builder/model driven forms part is done before merging in though.

seanmay:
Sounds good to me.
Wanted to make sure I wasn't heading in the wrong direction.

seanmay:
@e-schultz @barretodavid Would you mind having a look at the content updates?
Any objection to doing the model-driven content in a new feature-branch, or should I continue here?

handout/forms/nested-form-data.md
```diff
@@ -0,0 +1,94 @@
...
+    "provider": "Credit Lending Company Inc",
+    "cardholder": "Doug McKenzie",
+    "number": "123 456 789 012",
+    "verification": "321",
+    "expiry": "2020-02"
+  }
+}
+```
+
+You might have a form which has a 1-dimensional layout, with over-descriptive (or under-descriptive) names, and some controller or service to combine all of those fields into a data object like the above.
```
e-schultz: I understand what your trying to get at here, but it reads a little awkwardly.

e-schultz:
one part reads a little awkwardly, but otherwise I think this a good level of detail / examples. I'd say work on the model driven forms in this branch, and then merge it all at once to prevent a disconnect in the book between sections on new/old forms.

handout/forms/README.md
```diff
@@ -1,7 +1,7 @@
-# Form Builder
+# Forms
+An app without user any user input is really just a page. Capturing input from the user is the cornerstone of any application. In many cases, this means dealing with forms and all of their complexities.
 
-Capturing data from the user is the cornerstone of any application and it's usually done through forms. Angular 2 is much more flexible than Angular 1 for handling forms - we are no longer restricted to just the `ngModel`. In Angular 2, there are two ways to define forms: using directives in our templates, or using the `FormBuilder`.
+Angular 2 is much more flexible than Angular 1 for handling forms &emdash; we are no longer restricted to relying solely on `ngModel`. Instead, we are given degrees of simplicity and power, depending on what a form needs to accomplish.
```
barretodavid: There seems to be a typo: `&emdash`

handout/forms/getting-started.md
```diff
@@ -0,0 +1,40 @@
...
+```js
+import {bootstrap} from "@angular/core";
+// ...
+import {disableDeprecatedForms, provideForms} from "@angular/forms";
+
+bootstrap(MyApp, [
+  // ...
+  disableDeprecatedForms(),
+  provideForms()
+]);
```
barretodavid: I think it might be useful to explain why we need to call `disableDeprecatedForms()`. Angular still support both APIs and because it will automatically assign the `ngForm` directive to any `form` element, we need to make sure to deactivate the old api to avoid conflicts.

From [angular.io](https://angular.io/docs/ts/latest/guide/forms.html):
> Since we are converting to the new API, and no longer need the old API, we call disableDeprecatedForms() to disable the old form functionality and the warning message.

handout/forms/nested-form-data.md
```diff
@@ -0,0 +1,94 @@
...
+    <input type="email" name="contact_email" id="contact_email">
+
+    <label for="contact_phone">Phone</label>
+    <input type="text" name="contact_phone" id="contact_phone">
+  </fieldset>
+
+  <!-- ... -->
+
+</form>
+```
```
barretodavid: For the examples I would use shorter names to avoid visual clutter: `name="first-name"`

handout/forms/nested-form-data.md
```diff
@@ -0,0 +1,94 @@
...
+  <fieldset ngModelGroup="paymentCard">
+    <!-- ... -->
+  </fieldset>
+</form>
+```
+
+> - `id`s are only needed for HTML5 compatibility
+> - aside from semantic purposes, there is no reason `ngModelGroup` needs to be used on `<fieldset>` - it would work just as well on a `<div>`
+
+If we were to fill out the form, it would end up in the same shape as we need for our API, and we can still rely on the HTML field validation, if we know it‘s available.
```
barretodavid: I think two things should be mentioned here:
- Using `ngModelGroup` directive allow us to have block level validation which is specially useful for long forms.
- We should mention the relationship between `ngModelGroup`, `ngModel` and the property `name` of the fields. I have a general rule for this, every time we are introducing a new concept (or directive in this case) we should mention why is that needed, what problem is that solving.

handout/forms/template-driven_forms.md
```diff
@@ -0,0 +1,58 @@
...
+
+  <label for="password">Password</label>
+  <input type="password" name="password" id="password">
+
+  <button type="submit">Sign Up</button>
+</form>
+```
+
+The simplest method for dealing with forms in Angular 2 is not much more complicated than this.
+That's because Angular has already provided you a `form` directive (and friends), which operates under the covers.
```
barretodavid: It might be worth mentioning that Angular is actually implementing the `ngFor` directive to any `form` element found in a template. That will help you explain why the syntax `#signupForm="ngForm"` below. It's a way to export a reference to that particular instance of the `ngForm` directive.

handout/forms/template-model-binding.md
```diff
@@ -0,0 +1,74 @@
...
+  username: string = generateUniqueUserID();
+  email: string = '';
+
+  register(form: NgForm) {
+    console.log(form.value.username);
+    console.log(this.username);
+    // ...
+  }
+}
+```
```
barretodavid: There's also a third way: unidirectional data flow. In that case we use in the same `input` element the property `[ngModel]` and the event emitter `(ngModelChange)`. It's similar to 2 way data binding but you are in control on how to update the model.

handout/forms/README.md
```diff
@@ -1,7 +1,7 @@
...
 
-Capturing data from the user is the cornerstone of any application and it's usually done through forms. Angular 2 is much more flexible than Angular 1 for handling forms - we are no longer restricted to just the `ngModel`. In Angular 2, there are two ways to define forms: using directives in our templates, or using the `FormBuilder`.
+Angular 2 is much more flexible than Angular 1 for handling forms — we are no longer restricted to relying solely on `ngModel`. Instead, we are given degrees of simplicity and power, depending on what a form needs to accomplish.
 
-Using directives gives us the power of rapid prototyping without too much boilerplate, but we are somehow restricted in what we can do. The `FormBuilder`, on the other hand, lets us define our form through code and gives us much more flexibility and control over data validation.
-
-Which approach to use will depend on the programmer's needs, but we are going to start with the simplest one: directives.
+- [Template-Driven Forms](template-driven_forms.md) - Use built-in directives to create straightforward form components, with minimal code
```
e-schultz: links here don't work

handout/forms/form-builder/form-builder_basics.md
```diff
@@ -0,0 +1,67 @@
...
+})
+export class LoginForm {
+  loginForm: FormGroup;
+  username: FormControl;
+  password: FormControl;
+
+  constructor (builder: FormBuilder) {
+    this.username = new FormControl('', []);
+    this.password = new FormControl('', []);
+    this.testForm = builder.group({
```
e-schultz: should this be `this.loginForm` ?

handout/forms/form-builder/form-builder_basics.md
```diff
@@ -0,0 +1,67 @@
...
+  <label for="username">username</label>
+  <input type="text" name="username" id="username" [formControl]="username">
+
+  <label for="password">password</label>
+  <input type="password" name="password" id="password" [formControl]="password">
+
+  <button type="submit">log in</button>
+</form>
+```
+
```
e-schultz: Just created a plunker for this, https://plnkr.co/edit/PiCgcL?p=preview

handout/forms/form-builder/form-builder_validation.md
```diff
@@ -0,0 +1,64 @@
...
+  username: FormControl;
+  password: FormControl;
+
+  constructor (builder: FormBuilder) {
+    this.username = new FormControl('', [
+      Validators.required,
+      Validators.minLength(5)
+    ]);
+    this.password = new FormControl('', [Validators.required]);
+    this.testForm = builder.group({
```
e-schultz: should be changed to loginForm

handout/forms/form-builder/form-builder_validation.md
```diff
@@ -0,0 +1,64 @@
...
+      username: this.username,
+      password: this.password
+    });
+  }
+}
+```
+
+_app/login-form.component.html_
+```html
+<form [formGroup]="testForm" (ngSubmit)="submitForm()">
```
e-schultz: should be changed to loginForm

handout/forms/form-builder/form-builder_validation.md
```diff
@@ -0,0 +1,64 @@
...
+  </div>
+  <div >
+    <label for="password">password</label>
+    <input type="password" name="password" id="password" [formControl]="password">
+    <div [hidden]="password.valid || password.untouched">
+      <div>The following problems have been found with the password:</div>
+      <div [hidden]="!password.hasError('required')">The password is required.</div>
+    </div>
+  </div>
+  <button type="submit" [disabled]="!testForm.valid">Log In</button>
```
e-schultz: and here

handout/forms/form-builder/form-builder_validation.md
```diff
@@ -0,0 +1,64 @@
...
+_app/login-form.component.html_
+```html
+<form [formGroup]="testForm" (ngSubmit)="submitForm()">
+  Inside the form.
+  <div>
+    <label for="username">username</label>
+    <input type="text" name="username" id="username" [formControl]="username">
+    <div [hidden]="username.valid || username.untouched">
+      <div>The following problems have been found with the username:</div>
+      <div [hidden]="!username.hasError('minLength')">Username can not be shorter than 5 characters.</div>
```
e-schultz:
```html
 <div [hidden]="!username.hasError('minlength')">Username can not be shorter than 5 characters.</div>
```

small L

handout/forms/form-builder/form-builder_validation.md
```diff
@@ -0,0 +1,64 @@
...
+      Validators.required,
+      Validators.minLength(5)
+    ]);
+    this.password = new FormControl('', [Validators.required]);
+    this.testForm = builder.group({
+      username: this.username,
+      password: this.password
+    });
+  }
+}
```
e-schultz: I'd add the `login` handler here again to be consistent with the previous

handout/forms/form-builder/form-builder_validation.md
e-schultz: change from `submitForm` to `login` to be consistent with previous example

handout/forms/form-builder/form-builder_validation.md
```diff
@@ -0,0 +1,64 @@
...
+    <div [hidden]="password.valid || password.untouched">
+      <div>The following problems have been found with the password:</div>
+      <div [hidden]="!password.hasError('required')">The password is required.</div>
+    </div>
+  </div>
+  <button type="submit" [disabled]="!testForm.valid">Log In</button>
+</form>
+```
+
+Note that we have added rather robust validation on both the fields, and the form itself, using nothing more than built-in validators and some template logic.
```
e-schultz: created a plunker: https://plnkr.co/edit/NGejGL?p=preview

handout/forms/formbuilder/form-builder_custom_validation.md
```diff
@@ -0,0 +1,101 @@
...
+
+_app/login-form.component.ts_
+```html
+<!-- ... -->
+<div [hidden]="!password.hasError('needsExclamation')">
+  Your password must have an exclamation mark!
+</div>
+<!-- ... -->
+```
+
```
e-schultz: plunkr: https://plnkr.co/edit/t3Ut3P?p=preview

handout/forms/formbuilder/form-builder_custom_validation.md
```diff
@@ -0,0 +1,101 @@
...
+_app/login-form.component.html_
+```html
+<!-- ... -->
+<div [hidden]="!password.hasError('ampersandRequired')">
+  You must have an &amp; in your password.
+</div>
+<!-- ... -->
+```
+
```
e-schultz: plunker: https://plnkr.co/edit/9tWvhy?p=preview

handout/forms/form-builder/form-builder_basics.md
```diff
@@ -0,0 +1,67 @@
...
+### `FormControl`
+You will note that the `FormControl` class is assigned to similarly named fields, both on `this` and in the `FormBuilder#group({ })` method.
+This is mostly for ease of access.
+The form fields can otherwise be reached in the template by using `loginForm.controls.username` and `loginForm.controls.password`.
```
kosz: It would be good if we make sure this doesn't cause confusion and the users understand that they don't need to access a form's input by referencing the form, and instead, the user can just use the input name. This isn't clearly stated in this, or the paragraph above, but kind of visible in the source, but not clearly since there's no usage of the input field reference outside of the input ( ex: [hidden]="username.valid" ). 

this is also covered in later sections yes, but for this group of paragraphs, i feel some users may get the wrong idea. 

am i making any sense ?

handout/forms/form-builder/form-builder_basics.md
```diff
@@ -0,0 +1,67 @@
...
+Likewise, any instance of `FormControl` in this situation, can access its parent group by using its `.root` property (eg: `username.root.controls.password`).
+
+When constructing a `FormControl`, it takes two properties: an initial value, and a list of validators.
+Right now, we have no validation.
```
kosz: This ends abruptly, maybe it's a good place to briefly introduce the next section.

handout/forms/form-builder/form-builder_custom_validation.md
```diff
@@ -0,0 +1,102 @@
...
+Angular 2 allows you to do just that, with minimal effort.
+
+Let's assume we are using the same Login Form, but now we also want to test that our password has an exclamation mark somewhere in it.
+
+_app/login-form.component.ts_
+```ts
+function hasExclamationMark (input: FormControl) {
+  const hasExclamation = input.value.indexOf('!') >= 0;
+
+  return hasExclamation ? null : { needsExclamation: true };
```
kosz: ternary operators tend to confuse junior devs, am I alone in observing this or is it a well known fact ?
on the flip side, returning null explicitly makes a point that if validation passes you return null

handout/forms/form-builder/form-builder_custom_validation.md
```diff
@@ -0,0 +1,102 @@
...
+```html
+<!-- ... -->
+<div [hidden]="!password.hasError('needsExclamation')">
+  Your password must have an exclamation mark!
+</div>
+<!-- ... -->
+```
+
+### Predefined Parameters
+Having a custom validator check for exclamation marks might be helpful, but what if you need to check for some other form of punctuation?
```
kosz: this could also be called validator factories ? but that's also a highly technical term. 

the more important note I had here was that this chapter doesn't really talk about Angular, instead, it talks about using advanced javascript principles with Angular.  

I would anticipate that a user trying to grasp the concepts of Angular 2 forms may end up having to now spend energy and effort on trying to grasp these javascript features, if they are not familiar with them ( and angular 2 in a sense, is aimed at non js devs ). 

would it make sense to put this chapter somewhere further down ? 
would it also make sense to reconsider the name ? 

i'm not sure if these are "predefined parameters" to me these are javascript functions which can build new validators on the fly.
and since it's a more advanced topic, not really angular 2 specific, maybe the name can also reflect that ?

handout/forms/form-builder/form-builder_custom_validation.md
```diff
@@ -0,0 +1,102 @@
...
+It might be overkill to write nearly the same thing over and over again.
+
+Consider the earlier example `Validators.minLength(5)`.
+How did they get away with allowing you to pass in an argument to control the length?
+Simple, really.
+It's not a trick of Angular, or TypeScript.
+It's simple JavaScript closures.
```
kosz: will everyone reading this book know what a closure is ? i guess a better question is what type of audience are we targeting.

handout/forms/form-builder/form-builder_custom_validation.md
```diff
@@ -0,0 +1,102 @@
...
+_app/login-form.component.html_
+```html
+<!-- ... -->
+<div [hidden]="!password.hasError('ampersandRequired')">
+  You must have an &amp; in your password.
+</div>
+<!-- ... -->
+```
+
+### Validating Multiple Inputs
```
kosz: maybe this title can be misleading. 

we're not validating multiple inputs, we're validating ONE input while taking into consideration values of other inputs.

or we're running a custom validator, which needs to check the value of multiple input fields.

handout/forms/form-builder/form-builder_custom_validation.md
```diff
@@ -0,0 +1,102 @@
...
+Keep in mind what was mentioned earlier: inputs have access to their parent context via `.root`.
+Therefore, complex validation can happen by drilling through the form, via root.
+
+```ts
+function duplicatePassword (input: FormControl) {
```
e-schultz: had issues getting this to work, as input.root.controls has the potential to be undefined at a given point.

Working version:

```js
import {Component} from '@angular/core';
import {
  REACTIVE_FORM_DIRECTIVES,
  FormBuilder,
  FormControl,
  Validators
} from '@angular/forms';

function hasPunctuation (punctuation: string, errorType: string) {
  return function (input: FormControl) {
    return input.value.indexOf(punctuation) >= 0 ?
        null :
        { [errorType]: true };
  };
}

function duplicatePassword (input: FormControl) {
 
 if(!input.root.controls) {
   return null;
 }
 
  const exactMatch = input.root.controls.password.value === input.value;
  
  return exactMatch ? null : { mismatchedPassword: true };
}


@Component({
  selector: 'login-form',
  templateUrl: 'app/login-form.component.html',
  directives: [REACTIVE_FORM_DIRECTIVES]
})
export class LoginForm {
  loginForm: FormGroup;
  username: FormControl;
  password: FormControl;
  duplicatePassword: FormControl;
  
  constructor (builder: FormBuilder) {
    this.username = new FormControl('', [
      Validators.required,
      Validators.minLength(5)
    ]);
    
    this.password = new FormControl('', [Validators.required, hasPunctuation('&','ampersandRequired')]);
    this.duplicatePassword = new FormControl('', [
      Validators.required,
      duplicatePassword
    ]);
    
    this.loginForm = builder.group({
      username: this.username,
      password: this.password,
      duplicatePassword : this.duplicatePassword
    });
  }
  login () {
    console.log(this.loginForm.value);
    // Attempt Logging in...
  }
}
```

handout/forms/form-builder/form-builder_custom_validation.md
```diff
@@ -0,0 +1,102 @@
...
+  const exactMatch = input.root.controls.password = input.value;
+  return exactMatch ? null : { mismatchedPassword: true };
+}
+
+// ...
+this.duplicatePassword = new FormControl('', [
+  Validators.required,
+  duplicatePassword
+]);
+```
```
e-schultz: plunker: https://plnkr.co/edit/lMVp52?p=preview

handout/forms/form-builder/form-builder_validation.md
```diff
@@ -0,0 +1,64 @@
...
+
+We are using `.valid` and `.untouched` to determine whether or not we need to show errors. The field may be required, but if it hasn't been visited yet, we have not reason to tell the user that the value is wrong.
```
kosz: i think .valid, .untouched, .pristine, .touched deserve their own section. 
Optionally I'd also mention the ngMessages Angular 1 module here to indicate that Angular 2 is using the same type of mechanism to provide the ability to code in the best form UX possible.

handout/forms/form-builder/form-builder_custom_validation.md
```diff
@@ -0,0 +1,102 @@
+# FormBuilder Custom Validation
```
kosz: this file is duplicated ? there's a `form-builder` and a `formbuilder` directory

handout/forms/form-builder/form-builder_custom_validation.md
```diff
@@ -0,0 +1,102 @@
...
+function hasExclamationMark (input: FormControl) {
+  const hasExclamation = input.value.indexOf('!') >= 0;
+
+  return hasExclamation ? null : { needsExclamation: true };
+}
+
+
+// ...
+this.password = new FormControl('', [
+  Validators.required,
```
kosz: May be useful to have a short section covering some of the out of the box validators that angular 2 provides. I'm not sure if these are documented, but they can be seen in the source code :) 

https://github.com/angular/angular/blob/master/modules/%40angular/forms/src/validators.ts

I guess pattern is the one we didn't really cover. But having them in a clear list is also nice. 

Also interesting to note that pattern kind of does the same thing as the `!` example

handout/forms/form-builder/form-builder_custom_validation.md
e-schultz: some of the concepts are covered in our ES6/TypeScript section - we can't have every section targeting the most junior developers.  With the current course outline we have already covered Observables, Immutable and Dependency Injection.

handout/forms/form-builder/form-builder_validation.md
e-schultz: I think adding a section on styling forms where we talk about valid/pristine and the css-classes that angular adds is a good idea (and was there in our previous), but can be added in another PR and isn't necessary for tomorrow.

We barely mention Angular 1 in this course - and we have lots of people taking it that haven't been exposed to Angular 1, so think making mention to it isn't needed.

handout/forms/form-builder/form-builder_custom_validation.md
seanmay: Personally, I've found that the mental agility needed to get to "a = b || c" might trip people up momentarily, but that `function maybeY () { if (x) return y; }` as a structural flow will confuse people at a much less localized level.

I'm totally happy to have that chat, but I've dealt with a lot of misery down the other path, which ends up influencing architecture decisions all the way across the app, sometimes.

e-schultz: In this case, I'm fine with the ternary operator. In a case like this it's clear and concise, and in case someone asks "whats that?", its a quick aside conversation.

When I see stuff like

```
return  w ? (x ? 'yes' : 'what ') : y ? 'no' : 'huh'
```

... then I'd raise issues over readability/understandability;

handout/forms/form-builder/form-builder_custom_validation.md
e-schultz: agreed, but think it's fine if it's not there for tomorrow.

handout/forms/form-builder/form-builder_custom_validation.md
seanmay: I'm working on the assumption that there is less cognitive overhead to learn about a closure than to learn about developing a TypeScript interface expressly for passing in as a validator.

To wit, if we teach people that Validators are just functions, then what kind of TypeScript black-magic does a function-invocation (`Validators.minLength(5)`) return, such that the result of the function call can itself be used as a validator.
Don't we require the instantiation of a Noun to be the possessor of a verb? Is there some Angular injector we need?

No, it's really just a thunk.
There has to be some assumed level of JS proficiency, or at least higher-order function proficiency to get to the point of creating validators which hold closure references.
To that end, I would not expect junior programmers, who know nothing but Java-pre-8 to be able to create higher-order-functions nor custom validators which require them.

handout/forms/form-builder/form-builder_custom_validation.md
seanmay: Fair. I'll have a second take at that.

handout/forms/form-builder/form-builder_custom_validation.md
seanmay: Thanks for the catch.

handout/forms/form-builder/form-builder_custom_validation.md
seanmay: Yeah. IDE argument between renaming a folder, and saving a change in a buffer. Git source lost.

e-schultz:
I think this is good enough for round-1 of the new forms, @SethDavenport can give it a quick review tonight and merge if he feels good about talking to this content tomorrow

handout/forms/form-builder/form-builder.md
```diff
@@ -0,0 +1,9 @@
+# Form Builder
+
+Capturing data from the user is the cornerstone of any application and it's usually done through forms. Angular 2 is much more flexible than Angular 1 for handling forms - we are no longer restricted to just the `ngModel`. In Angular 2, there are two ways to define forms: using directives in our templates, or using the `FormBuilder`.
+
```
SethDavenport: This text seems familiar :)  Let's remove the duplication in a second pass.

handout/forms/form-builder/form-builder_basics.md
SethDavenport: Thanks for this @e-schultz

SethDavenport:
Thanks @seanmay, @e-schultz, and @kosz for getting this done. I'll take it for a spin tomorrow and report back :)

PR #626, opened 2016-07-20
Added a couple of ngrx plunkrs.

SethDavenport:
Also adjusted the 'getting more from redux & ngrx section'


SethDavenport:
Was asked about ngrx in todays course - don't have time to produce full content before tomorrow so I'll cover the ng2-redux material and compare/contrast with ngrx. Having both plunkrs will help.

Turns out they're still very similar to use. Still like `@select` a lot more tho.

SethDavenport:
@e-schultz @seanmay merging since I'm teaching it tomorrow. a _post-hoc_ review would be appreciated.

PR #632, opened 2016-07-21
Feature template forms

seanmay:
Added Visual Cues, added template validation, corrected problems in example code.


handout/forms/template-driven/validating_forms.md
```diff
@@ -0,0 +1,31 @@
+# Validating Template-Driven Forms
+
+
+### Validation
+Using the Template-Driven approach, form validation is a matter of following HTML5 practices.
+
+```html
+<!-- a required field -->
+<input type="text" ... required>
```
bennett000: We _might_ want to change this to:

```
<input type="text" <!-- implementation --> required>
```

or something else so that people like me aren't accidentally mis-interpreting ellipses for destructuring

I might just be crazy though second opinions would be good.  I'm suggesting this not-just because of my own overly literal interpretation but because of some of the comments I've seen on our GitBook

bennett000:
The update itself LGTM.  I was under the impression the changes to forms were a lot more significant... were there other edits to forms over the last week that I'm behind on?

PR #648, opened 2016-08-04
LL-Update template-driven_forms.md

llittlewtw:
copy edits
@SethDavenport I don't know what the "(and friends)" refers to.


e-schultz:
@seanmay @SethDavenport @llittlewtw I think Sean is referring to common HTML form directives that angular has built in.

llittlewtw:
@e-schultz @seanmay @SethDavenport  Is "friends" understood to have that meaning in the JS dev lingo? Leave as is?

PR #654, opened 2016-08-04
LL-Update form-builder_custom_validation.md

llittlewtw:
copy edits
@SethDavenport  - "How did they get away with..."  - Who is "they"? Is it OK/accurate to change to "Why was an argument to control the length allowed, if a validator is just a function?"


e-schultz:
@llittlewtw read your edits and it makes sense, 👍  thanks

PR #661, opened 2016-08-05
LL-Update routerlink.md

llittlewtw:
copy edits

### DIFF
--- a/handout/routing/routerlink.md
+++ b/handout/routing/routerlink.md
@@ -2,7 +2,7 @@
 
 ## RouterLink ##
 
-Links to routes can be added using the `RouterLink` directive, which is part of `ROUTER_DIRECTIVES` along with `RouterOutlet`.
+Add links to routes using the `RouterLink` directive, which is part of `ROUTER_DIRECTIVES` along with `RouterOutlet`.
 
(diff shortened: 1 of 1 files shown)


handout/routing/routerlink.md
```diff
@@ -2,7 +2,7 @@
 
 ## RouterLink ##
 
-Links to routes can be added using the `RouterLink` directive, which is part of `ROUTER_DIRECTIVES` along with `RouterOutlet`.
+Add links to routes using the `RouterLink` directive, which is part of `ROUTER_DIRECTIVES` along with `RouterOutlet`.
```
e-schultz: Trying to think of how to best explain this. To me, this reads like we are creating new routes, but what the RouterLink does is create a link to an existing route.

llittlewtw:
@e-schultz  It says "add links to routes", but I can see that some might infer that that might include creating a new route at the same time. Sufficient to just add "existing"?

llittlewtw:
@e-schultz We're trying to avoid passive language construction. The active language version of what you wrote is "create links to routes" which doesn't address your initial concern about interpreting the statement as creating new routes. You don't like "Add links to existing routes..." ?

Issue #670, opened 2016-08-15
Update Bootstrap section to use `@NgModule`


e-schultz:
Bootstrapping section is using old RC5 style bootstrapping.

Should update to cover the minimum needed for `@NgModule` bootstrapping the application

PR #685, opened 2016-08-19
Added the modules section and updated gitbook to version 3+

barretodavid:
connects to #670


handout/modules/introduction.md
```diff
@@ -0,0 +1,89 @@
+# What is an Angular 2 Module?
+
+In Angular 2, a module is a mechanism to group components, directives, pipes and services that are related, in such a way that can be combined with other modules to create an application. An Angular 2 application can be thought of as a puzzle where each piece (or each module) is needed to be able to see the full picture.
+
+Another analogy to understand Angular 2 modules is classes. In a class, we can define public or private methods. The public methods are the API that other parts of our code can use to interact with it while the private methods are implementation details that are hidden. In the same way, a module can export or hide components, directives, pipes and services. The exported elements are meant to be used by other modules, while the ones that are not exported (hidden) are just used inside the module itself and cannot be directly accessed by other modules of our application.
+
+## A Basic Use of Modules
+
+To be able to define modules we have to use the decorator `NgModule`.
+
```
smithad15: Add `app/app.module.ts` file name

handout/modules/introduction.md
```diff
@@ -0,0 +1,89 @@
...
+  imports: [ ... ],
+  declarations: [ ... ],
+  bootstrap: [ ... ]
+})
+export class AppModule { }
+```
+
+In the example above, we have turned the class `AppModule` into an Angular 2 module just by using the `NgModule` decorator. The `NgModule` decorator requires at least three properties: `imports`, `declarations` and `bootstrap`.
+
+The property `imports` expects an array of modules. Here's where we define the pieces of our puzzle (our application). The property `declarations` expects an array of components and pipes that are part of the module. The `bootstrap` property is where we define the root component of our module. Even though this property is also an array, 99% of the time we are going to define only one component.
```
smithad15: `declarations` expects an array of components, _directives_ and pipes

handout/modules/introduction.md
```diff
@@ -0,0 +1,89 @@
...
+  bootstrap: [AppComponent]
+})
+export class AppModule { }
+```
+
+The file _app.component.ts_ is just a "hello world" component, nothing interesting there. In the other hand, the file _app.module.ts_ is following the structure that we've seen before for defining a module but in this case, we are defining the modules and components that we are going to be using.
+
+The first thing that we notice is that our module is importing the `BrowserModule` as an explicit dependency. The `BrowserModule` is a built-in module that exports basic directives, pipes and services like `NgFor` and `NgIf`. Unlike previous versions of Angular 2, we have to explicitly import those dependencies to be able to use directives like `*ngFor` or `*ngIf` in our templates.
+
+Given that the root (and only) component of our module is the `AppComponent` we have to list it in the `bootstrap` array. Because in the `declarations` property we are supposed to defined **all** the components or pipes that make up our application, we have to define again the `AppComponent` there.
```
smithad15: "...in the declarations property we are supposed to _define_ ..."
"...we have to define the `AppComponent` again there too."

handout/modules/introduction.md
```diff
@@ -0,0 +1,89 @@
...
+
+Given that the root (and only) component of our module is the `AppComponent` we have to list it in the `bootstrap` array. Because in the `declarations` property we are supposed to defined **all** the components or pipes that make up our application, we have to define again the `AppComponent` there. 
+
+Before moving on, there's an important clarification to make. **There are two types of modules, root modules and feature modules**.
+
+In the same way that in a module we have one root component and possible many secondary components, **in an application we only have one root module and zero or many feature modules**. To be able to bootstrap our application, Angular needs to know which one is the root module. An easy way to identify a root module is by looking at the `imports` property of its `NgModule` decorator. If the module is importing the `BrowserModule` then it's a root module, if instead is importing the `CommonModule` then is a feature module.
```
smithad15: "...and many possible secondary components..."
"...importing the `CommonModule` then _it_ is a feature module."

handout/modules/introduction.md
```diff
@@ -0,0 +1,89 @@
...
+
+> As developers, we need to take care of **only** importing the `BrowserModule` in the root module and using the `CommonModule` in any other module we create for the same application. Failing to do so might result in problems when working with lazy loaded modules as we are going to see in following sections.
```
smithad15: Word order makes it seem like we shouldn't be importing anything else but `BrowserModule` into root modules

handout/modules/introduction.md
```diff
@@ -0,0 +1,89 @@
...
+
+By convention, the root module should always be named `AppModule`.
+
+## Bootstraping an Application
```
smithad15: "Bootstrapping"

handout/modules/multiple-elements.md
```diff
@@ -0,0 +1,114 @@
...
+    AppComponent,
+    CreditCardMaskPipe,
+    CreditCardComponent
+  ],
+  bootstrap: [AppComponent]
+})
+export class AppModule { }
+```
+
+Notice that we have added the component `CreditCardComponent` and the pipe `CreditCardMaskPipe` to the `declarations` property, along with the root component of the module `AppComponent`. In the other hand, our custom service is defined in another property of called `providers`.
```
smithad15: "Our custom service is configured with the dependency injection system with the `providers` property."

handout/modules/feature-modules.md
```diff
@@ -0,0 +1,157 @@
+# Creating a Feature Module
+
+When our root module start growing, it start to be evident that some elements (components, directives, etc.) are related in a way that almost feel like they belong to a library that can be "plugged in".
```
smithad15: "When our root module starts growing, it starts..."

handout/modules/introduction.md
barretodavid: @smithad15 I didn't put the filename there because that snippet was mostly to give a an overview of how the `NgModule` decorator looks like. I'm "formally" starting the code example in the next snippet and that's when I put the filename so people can code along if they want to.

winkerVSbecks:
:shipit: We should add a link to the lazy load section once that is done.

barretodavid:
@winkerVSbecks I was thinking that the lazy loaded section belongs to the router chapter. What do you think? Should that be included in the NgModule chapter instead?

winkerVSbecks:
@barretodavid I meant once we add the lazy loading section to the router chapter we should add a link to it in `feature-modules.md`

Issue #522, opened 2016-06-27
Component interaction with local template variable


e-schultz:
* [x] - Finalize outline
* [x] - Content
* [x] - Example
* [ ] - Exercise 

Show effective use/use cases for this.
```html
 <some-compoent #temp></some-component> {{#temp.something}}
```

gillianml:
@e-schultz would you please review and merge the PR if it's good?

PR #690, opened 2016-08-23
Added section on Template Reference Variables connects to #522

seanmay:
Added section on Template Reference Variables

### DIFF
--- a/handout/components/app_structure/access_child_components_from_template.md
+++ b/handout/components/app_structure/access_child_components_from_template.md
@@ -1,0 +1,38 @@
+# Access Child Components From the Template
+In our templates, we may find ourselves needing to access values provided by the child components which we use to build our own component.
+
+The most straightforward examples of this may be seen dealing with forms or inputs:
+
+_app/my-example.component.html_
+```html
+<section >
+  <form #myForm="ngForm" (ngSubmit)="submitForm(myForm)">
+    <!-- ... -->
+  </form>
+</section>
+```
+
+
+_app/my-example.component.ts_
+```ts
+// ...
+import {NgForm} from '@angular/forms';
+// ...
+export class MyExampleComponent {
+  submitForm (form: NgForm) {
+    // ...
+  }
+}
+```
+[View Example](https://plnkr.co/edit/MFaLMCbLOIXfNhYWT9ni?p=preview)
+
+This isn't just a magic feature which only forms have.
+
+```html
+<red-ball #myBall></red-ball>
+The ball is {{ myBall.color }}.
+```
+
+By assigning a Template Reference Variable to a component, you can read properties from it, call methods on it, or pass the instance into your own method calls (as the form was passed in, above).
+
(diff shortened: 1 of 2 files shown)


handout/components/app_structure/access_child_components_from_template.md
```diff
@@ -0,0 +1,38 @@
...
+    // ...
+  }
+}
+```
+[View Example](https://plnkr.co/edit/MFaLMCbLOIXfNhYWT9ni?p=preview)
+
+This isn't just a magic feature which only forms have.
+
+```html
+<red-ball #myBall></red-ball>
```
winkerVSbecks: Wasn't really sure where the value for `myBall` is coming from so, made a plunker for it: https://plnkr.co/edit/UTBleHSpTf5To8zZSbSd?p=preview

IMO it's probably worth adding that in or something similar

winkerVSbecks:
@seanmay apart from that one minor suggestion :shipit:

handout/components/app_structure/access_child_components_from_template.md
```diff
@@ -0,0 +1,38 @@
...
+The ball is {{ myBall.color }}.
+```
+
+By assigning a Template Reference Variable to a component, you can read properties from it, call methods on it, or pass the instance into your own method calls (as the form was passed in, above).
+
+There are better means of dealing with complex tasks, which might require you to convey information to the child, or interact with its interface; however, if you find yourself simply needing to reference properties extended by the child, for use in your template, or to pass a value from a child to your method, in an action, this is a straightforward approach.
```
e-schultz: I'm not sure if 'better ways' is the best way of saying it - as I think when combined with projection, can actually simplify things quite a bit. ie) can reference variables in the template without the 'container' needing to know what is being projected into it. 

An important thing to note - is this doesn't give you access to those properties within the backing class of the component. To do that, you would need to use ContentChildren or ViewChildren to query the components (and think we have a section on that). 

An example I was working on for something was along the lines of

```html
<some-container #stuff [somedata]="inputs" (event)="handleEvent($event)">
<button (click)="doSomethingWithStuffsData(stuff.data)">Save!</button>
</some-container>
```

The public properties/methods on your components can act as a bit of an interface to allow for interaction with others. 

In the above example, `some-container` doesn't need to know anything about the thing that is making use of it's stuff - you don't need to deal with querying children/view children or do any extra lifting on `some-container`.

handout/components/app_structure/access_child_components_from_template.md
```diff
@@ -0,0 +1,38 @@
+# Access Child Components From the Template
+In our templates, we may find ourselves needing to access values provided by the child components which we use to build our own component.
```
barretodavid: The first time that I read this paragraph I thought you were going to show the `@Output` decorator. I think it would be better to talk about "component's instance" instead of "component's values" when referring to template variables.

What about something about these lines:

_Oftentimes, we may find ourselves needing to access an instance of our child components which we use in other parts of our template._

e-schultz:
@seanmay do you have the capacity to address the comments, or could/should someone take over the final edits? fine either way

seanmay:
I can fix these post-interview in 30 minutes.

e-schultz:
:shipit:

Issue #672, opened 2016-08-15
Update Testing section to cover RC5


e-schultz:
* RC5 introduced a few breaking changes to unit testing
* Also includes features for working with Modules
* [Changelog](https://github.com/angular/angular/blob/master/CHANGELOG.md#breaking-changes)

PR #693, opened 2016-08-25
Updated the testing section to RC5

winkerVSbecks:
- Updated all code snippets to RC5 syntax
- Added plunker links for all examples
- Restructured the service testing section a bit and added an alternative http mock strategy

connects to #672


handout/testing/components/injecting-dependencies.md
```diff
@@ -24,71 +24,71 @@ export class QuoteComponent {
...
-  TestComponentBuilder,
-  ComponentFixture,
+class MockQuoteService {
+  public quote: string = 'Test quote';
+
+  getQuote() {
+    return new Promise<string>((resolve, reject) => {
+      resolve(this.quote);
+    });
+  }
```
barretodavid: What if we simplify the method by just doing:

```js
getQuote() {
  return Promise.resolve(this.quote);
}
```

handout/testing/components/injecting-dependencies.md
```diff
@@ -24,71 +24,71 @@ export class QuoteComponent {
...
+    fixture.detectChanges();
+  });
+
+  it('Should get quote', async(inject([], () => {
+    fixture.componentInstance.getQuote();
+    fixture.whenStable().then(() => {
+      fixture.detectChanges();
+      fixture.whenStable().then(() => {
+        const compiled = fixture.debugElement.nativeElement;
+        expect(compiled.querySelector('div').innerText).toEqual('Test quote');
```
barretodavid: Can we chain promises here? Something like:

```js
fixture.whenStable()
  .then(() => {
    fixture.detectChanges();
    return fixture.whenStable();
  })
  .then(() => {
    const compiled = fixture.debugElement.nativeElement;
    expect(compiled.querySelector('div').innerText).toEqual('Test quote');
  });
```

winkerVSbecks: That's so much better. Thanks

winkerVSbecks:
@barretodavid updated PR with changes you suggested.

handout/testing/components/async.md
```diff
@@ -1,43 +1,78 @@
...
+class MockQuoteService {
+  public quote: string = 'Test quote';
 
-      fixture.detectChanges();
-      expect(fixture.debugElement.nativeElement.querySelector('span'))
-        .toHaveText('My expected data');
+  getQuote() {
+    return new Promise<string>((resolve, reject) => {
+      resolve(this.quote);
     });
```
barretodavid: We could use the same simplification here of just using `return Promise.resolve(this.quote);`.

handout/testing/services/injecting-dependencies.md
```diff
@@ -1,3 +1,3 @@
 # Injecting Dependencies
 
-Like components, services often require dependencies that Angular injects through the constructor of the service's class. Since we are initializing these classes outside the bootstrapping process of Angular, we must explicitly inject these dependencies ourselves. This is accomplished by calling `addProviders` in the `beforeEach` part of your unit test to feed in required dependencies like the HTTP module.
+Like components, services often require dependencies that Angular injects through the constructor of the service's class. Since we are initializing these classes outside the bootstrapping process of Angular, we must explicitly inject these dependencies ourselves. This is accomplished by using the `TestBed` to configure a testing module and feed in required dependencies like the HTTP module.
```
barretodavid: I was reading at the book and I noticed that there are two consecutive chapters that only have one paragraph: "Testing Strategies for Services" and "Injecting Dependencies". How do you feel about merging them in just one chapter?

handout/testing/services/mockbackend.md
```diff
@@ -0,0 +1,103 @@
...
+        SearchWiki
+      ]
+    });
+  });
+
+  it('should get search results', fakeAsync(
+    inject([
+      XHRBackend,
+      SearchWiki
+    ], (mockBackend, searchWiki) => {
```
barretodavid: Is Typescript able to infer the types for `mockBackend` and `searchWiki`?

winkerVSbecks: I'm not sure. Should we provide type annotation for them?

handout/testing/services/mockbackend.md
```diff
@@ -0,0 +1,103 @@
...
+
+});
+```
+[View Example](http://plnkr.co/edit/k5bQN2YuFMkNXrNC2Zo3?p=preview)
+
+We use `inject` to inject the `SearchWiki` service and the `MockBackend` into our test. We then wrap our entire test with a call to `fakeAsync`, which will be used to control the asynchronous behavior of the `SearchWiki` service for testing.
+
+Next, we `subscribe` to any incoming connections from our back-end. This gives us access to an object `MockConnection`, which allows us to configure the response we want to send out from our back-end, as well as test any incoming requests from the service we're testing.
+
+In our example, we want to verify that the `SearchWiki`'s search method makes a GET request to the correct URL. This is accomplished by looking at the request object we get when our `SearchWiki` service makes a connection to our mock back-end. Analyzing the `request.url` property we can see if its value is what we expect it to be. Here we are only checking the URL, but in other scenarios we can see if certain headers have been set, or if certain POST data has been sent.
```
barretodavid: > we want to verify that the `SearchWiki`'s search method makes a GET request to the correct URL

In the test file we are verifying the request URL but we are not checking the method. Maybe we should add an additional expectation there to be consistent:

```js
expect(connection.request.method).toBe('GET');
expect(connection.request.url).toBe(expectedUrl);
```

handout/testing/services/mockbackend.md
```diff
@@ -0,0 +1,103 @@
...
+
+It is possible to use multiple responses. Say your service had two possible GET requests - one for `/api/users`, and another `/api/users/1`. Each of these requests has a different corresponding set of mock data. When receiving a new connection through the `MockBackend` subscription, you can check to see what type of URL is being requested and respond with whatever set of mock data makes sense.
+
+Finally, we can test the `search` method of the `SearchWiki` service by calling it and subscribing to the result. Notice that we call `tick` before checking the contents of the search result. This is to ensure that all `Observable`s, `Promise`s, etc. and any other asynchronous operations complete before proceeding. Once our search process has finished, we check the result object to see if it contains the same data that we mocked into our back-end. If it is, then congratulations, your test has passed.
```
barretodavid: I think the `tick` function is missing from the example code.

barretodavid: Also, I think the `tick` function should be called after the subscription, not before as the text implies don't you think?

```js
searchWiki.search('Angular')
  .subscribe(res => {
    expect(res).toEqual(mockResponse);
  });

tick();
```

handout/testing/services/alt-http-mocking.md
```diff
@@ -0,0 +1,83 @@
...
+
+  beforeEach(() => {
+    mockHttp = { get: null } as Http;
+
+    spyOn(mockHttp, 'get').and.returnValue(Observable.of({
+      json: () => mockResponse
+    }));
+
+    TestBed.configureTestingModule({
+      imports: [HttpModule],
```
barretodavid: Why do we even have to import the `HttpModule` if we are mocking the `Http` service directly in the `TestModule`?

barretodavid: I just tried removing that line (`imports: [HttpModule]`) in the plunker and it worked :)

handout/testing/services/mockbackend.md
winkerVSbecks: Interesting thing is I don't even have `tick` in the actual plunker 😱 But, I'll add one to the end

handout/testing/services/alt-http-mocking.md
winkerVSbecks: That's strange. It doesn't work for me without that. I get `Error: No provider for Http!`

handout/testing/services/mockbackend.md
winkerVSbecks: I guess we don't need tick here since we are not using a timeout. Should I leave it out or use a different example?

winkerVSbecks: Updated the plunker with another test that uses timeout. And tick needs to come before the assertion. 

```
  it('should get search results', fakeAsync(
    inject([SearchWiki], (searchWiki) => {
      searchWiki.setFoo();
      tick(1000);
      expect(searchWiki.foo).toEqual('food');
    })
  ));
```

handout/testing/services/mockbackend.md
barretodavid: I think so, otherwise we will not get code completion from the IDE.

```js
inject(
  [XHRBackend, SearchWiki], 
  (mockBackend: MockBackend, searchWiki: SearchWiki) => {
    ...
  }
)
```

handout/testing/services/alt-http-mocking.md
```diff
@@ -0,0 +1,83 @@
+# Alternative HTTP Mocking Strategy
+
+An alternative to using `MockBackend` is to create light mocks by ourselves. Here we create an object and then tell TypeScript to treat it as `Http` using type assertion. We then create a spy for its `get` method and return an observable similar to what the real `Http` service would do.
```
e-schultz: How about:

An alternative to using `MockBackend` is to create our own light mocks ~~by ourselves~~.

handout/testing/services/alt-http-mocking.md
```diff
@@ -0,0 +1,83 @@
+# Alternative HTTP Mocking Strategy
```
e-schultz: @SethDavenport @winkerVSbecks possible blog-post idea 'non-insane HTTP testing in Angular 2' ? :P

handout/testing/services/mockbackend.md
```diff
@@ -0,0 +1,115 @@
+# Testing HTTP Requests Using MockBackend
+
```
e-schultz: This seems a little wordy, wondering if something like

> To unit test our services, we don't want to make actual HTTP requests. To accomplish this, we need to mock out our HTTP services. Angular 2 provides us with a `MockBackend` class that can be configured to provide mock responses to our requests, without actually making a network request. 
>
> The configured `MockBackend` can then be injected into HTTP, so any calls to the service, such as `http.get` will return our expected data, allowing us to test our service in isolation from real network traffic.

is any better? ... might need some tweaks still - just feels like the last part of the first paragraph is largely repeating the first part of the second.

winkerVSbecks: Much better

e-schultz:
👍 looking good to me, unless @barretodavid has any other suggestions fine to :shipit:

Issue #675, opened 2016-08-15
Cover Aux Routes in Routing


e-schultz:
* RC5 / New router release added support for Aux routes in the `routerLink` DSL 
* Add/Update section to cover Aux routes

PR #695, opened 2016-08-26
added the aux routes chapter

winkerVSbecks:
The content is pretty much the same as what we had for the deprecated router but, with an updated example. 

connects to #675


e-schultz:
:shipit:

PR #707, opened 2016-08-30
Updates Advanced Components section to RC5

rpdasilva:
connects to #704


e-schultz:
@rpdasilva there are copies of these examples in the repo, ie, `examples/advanced-components/...` - could those also get updated to RC5? Plunkr can be finicky at times and have had to goto the copies in git when doing live coding during training.

rpdasilva:
@e-schultz Updated

rpdasilva:
closes #704

PR #719, opened 2016-09-01
update component examples, closes #703

### LINKED ISSUE #703: Update code examples: Components in Angular 2

kosz:


bennett000:
I think the CLI section is good enough for RC5 and the current (broken/transitionary) state of Angular CLI itself.

In the future we'll want to talk about generating modules and other things, but for now :shipit:

PR #724, opened 2016-09-06
Basic upgrade of DI to RC6 and NgModule

### LINKED ISSUE #674: Update DI section to cover RC5/NgModule
Things to cover:
* providers in an `@NgModule` get application scoped`
* the `provide` method is getting deprecated, and should be using the object literal instead.

e.g) in 'injection beyond classes',

```js
// before

provide(Hamburger, { useClass: Hamburger })

// after

{ provide: Hamburger, useClass: Hamburger }
```

* content in section should move away from `bootstrap` and to using `@NgModule`
* examples should be updated to use `@NgModule`
* While we are at it, update examples to follow [angular.io styleguide](https://angular.io/styleguide) for naming conventions/structure

bennett000:
Fixes #674 

Upgrades DI section to use `@NgModule` and modern provider Objects.  This is the basic upgrade.

What would be nice expansions

- Opaque tokens
- Basic multiple module example (with shared dependencies)
- Advanced multiple module example (with dependency conflicts and namespace issues)
- Lazy loading example

Lazy loading is likely out of scope for this PR, but the first three could be addressed here, or would new cards be better?


rpdasilva:
LGTM :shipit:

Regarding your points.
It seems to me that they each probably deserve their own cards. A case could be argued to include OpaqueTokens since it's really just a tiny addition, but given how bite-sized some of these handouts have been - I could see one handout expanding on replacing the string token 'Random' with OpaqueToken 'Random', the reasoning why, and another plnkr. (which would require subsequent plnkr/code example updates to use said OpaqueToken)

PR #728, opened 2016-09-06
RC6 Components

rpdasilva:
connects to #725


bennett000:
This looks good to me although there's a lot of edits it looks like what the label says on the tin.  :shipit:

Issue #735, opened 2016-09-07
DI: Lazy loading


bennett000:
Add a section to DI that discusses the implications of lazy loading with respect to DI and singletons.

PR #754, opened 2016-09-16
Added sections for Lazy Loading and DI

barretodavid:
Connects to #735

### DIFF
--- a/handout/modules/lazy-load-di.md
+++ b/handout/modules/lazy-load-di.md
@@ -1,0 +1,93 @@
+# Lazy Loading and the Dependency Injection Tree
+
+Lazy loaded modules create their own branch on the Dependency Injection (DI) tree. This means that it's possible to have services that belong to a lazy loaded module, that are not accessible by the root module or any other eagerly loaded module of our application.
+
+To show this behaviour, let's continue with the example of the previous section and add a `CounterService` to our `LazyModule`.
+
+_app/lazy/lazy.module.ts_
+
+```js
+...
+import { CounterService } from './counter.service';
+
+@NgModule({
+  ...
+  providers: [CounterService]
+})
+export class LazyModule {}
+```
+
+Here we have added the `CounterService` to the `providers` array. Our `CounterService` is a simple class that holds a reference to a `counter` property.
+
+_app/lazy/counter.service.ts_
+
+```js
+import { Injectable } from '@angular/core';
+
+@Injectable()
+export class CounterService {
+  counter = 0;
+}
+```
+
+We can now modify the `LazyComponent` to use this service with a button to increment the `counter` property.
+
+_app/lazy/lazy.component.ts_
+
+```js
+import { Component } from '@angular/core';
+
+import { CounterService } from './counter.service';
+
+@Component({
+  template: `
+    <p>Lazy Component</p>
+    <button (click)="increaseCounter()">Increase Counter</button>
+    <p>Counter: {{ counterService.counter }}</p>
+  `
+})
+export class LazyComponent {
+
+  constructor(public counterService: CounterService) {}
+
+  increaseCounter() {
+    this.counterService.counter += 1;
+  }
+}
+```
+
+[View Example](https://plnkr.co/edit/RUp3QhHWmxBIQQAAw2im?p=preview)
+
+The service is working. If we increment the counter and then navigate back and forth between the `eager` and the `lazy` routes, the `counter` value will persist in the lazy loaded module. 
+
+But the question is, how can we verify that the service is isolated and cannot be used in a component that belongs to a different module? Let's try to use the same service in the `EagerComponent`.
+
+_app/eager.component.ts_
+
+```js
+import { Component } from '@angular/core';
+import { CounterService } from './lazy/counter.service';
+
+@Component({
+  template: `
+    <p>Eager Component</p>
+    <button (click)="increaseCounter()">Increase Counter</button>
+    <p>Counter: {{ counterService.counter }}</p>
+  `
+})
+export class EagerComponent {
+  constructor(public counterService: CounterService) {}
+
+  increaseCounter() {
+    this.counterService.counter += 1;
+  }
+}
+```
+
+If we try to run this new version of our code, we are going to get an error message in the browser console:
+
+```
+No provider for CounterService!
+```
+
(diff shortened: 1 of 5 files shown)


handout/modules/lazy-load-di.md
```diff
@@ -0,0 +1,93 @@
...
+```js
+import { Injectable } from '@angular/core';
+
+@Injectable()
+export class CounterService {
+  counter = 0;
+}
+```
+
+We can now modify the `LazyComponent` to use this service with a button to increment the `counter` property.
```
bennett000: drop the `now` (style guide active voice) from `We can now modify`

handout/modules/lazy-load-di.md
```diff
@@ -0,0 +1,93 @@
...
+import { CounterService } from './counter.service';
+
+@NgModule({
+  ...
+  providers: [CounterService]
+})
+export class LazyModule {}
+```
+
+Here we have added the `CounterService` to the `providers` array. Our `CounterService` is a simple class that holds a reference to a `counter` property.
```
bennett000: drop the `have` from `Here we have added` (style guide active voice)

handout/modules/lazy-load-di.md
```diff
@@ -0,0 +1,93 @@
...
+}
+```
+
+If we try to run this new version of our code, we are going to get an error message in the browser console:
+
+```
+No provider for CounterService!
+```
+
+What this error is telling us is that the `AppModule`, where the `EagerComponent` is defined, has no knowledge of a service called `CounterService`. `CounterService` lives in a different branch of the DI tree created for `LazyModule` when it was lazy loaded in the browser.
```
bennett000: Rephrase `What this error is telling us` to `What this error tells us`

handout/modules/lazy-loading-module.md
```diff
@@ -0,0 +1,152 @@
+# Lazy Loading a Module
+
+Another advantage of using modules to group related pieces of functionality of our application, is the ability to load those pieces on demand. Lazy loading modules will help us decrease the startup time of our application because we don't need to load everything at once, we only need to load what the user expects to see at the beginning and then conditionally load other modules as the user navigates to different sections. As you can imagine, there's a relationship between lazy loading and routing.
```
bennett000: - remove the comma from `functionality of our application, is the`
- change `Lazy loading modules will help us decrease` to `Lazy loading modules helps us decrease`
- break the sentence `Lazy loading modules will help us decrease the startup time of our application because we don't need to load everything at once, we only need to load what the user expects to see at the beginning and then conditionally load other modules as the user navigates to different sections.` into `...decrease the startup time.  With lazy loading our application does not need to load everything at once, it only needs to load what the user expects to see when the app first loads.  Modules that are lazily loaded will only be loaded when the user navigates to their routes.`

handout/modules/lazy-loading-module.md
```diff
@@ -0,0 +1,152 @@
...
+```js
+import { Component } from '@angular/core';
+
+@Component({
+  template: '<p>Eager Component</p>'
+})
+export class EagerComponent {}
+```
+
+But more importantly, we can see that whenever we try to go to the path `lazy`, we are going to lazy load a module called conveniently `LazyModule`. Look closely at the definition of that route:
```
bennett000: `called conveniently` to `conveniently called`

handout/modules/lazy-loading-module.md
```diff
@@ -0,0 +1,152 @@
...
+
+```js
+{ path: 'lazy', loadChildren: 'lazy/lazy.module#LazyModule' }
+```
+
+There's a few important things to notice here:
+
+1. We are using the property `loadChildren` instead of `component`.
```
bennett000: `We are using` to `We use` (style guide active voice)

handout/modules/lazy-loading-module.md
```diff
@@ -0,0 +1,152 @@
...
+2. We are passing a string instead of a symbol to avoid loading the module eagerly.
```
bennett000: - `We are passing` to `We pass` (style guide active voice)

handout/modules/lazy-loading-module.md
```diff
@@ -0,0 +1,152 @@
...
+3. We are defining not only the path to the module but the name of the class as well.
```
bennett000: `We are defining` to `We define` (style guide active voice)

handout/modules/lazy-loading-module.md
```diff
@@ -0,0 +1,152 @@
...
+import { LazyComponent } from './lazy.component';
+
+const routes: Routes = [
+  { path: '', component: LazyComponent }
+];
+
+export const routing: ModuleWithProviders = RouterModule.forChild(routes);
+```
+
+Notice that we are using the method call `forChild` instead of `forRoot` to create the routing object. We should always do that when creating a routing object for a feature module, no matter if the module is supposed to be eagerly or lazy loaded.
```
bennett000: - `we are using` to `we use` (style guide active voice)
- `no matter if the module is supposed to be eagerly or lazy loaded` to `no matter if the module is supposed to be loaded eagerly or lazily`

handout/modules/shared-di-tree.md
```diff
@@ -0,0 +1,62 @@
+# Sharing the Same Dependency Injection Tree
+
+So far our problem is that we are creating two instances of the same services in different levels of the DI tree and that the instance that is created in the lower branch of the tree, is shadowing the one defined at the root level. The solution? To avoid creating a second instance in a lower level of the DI tree for the lazy loaded module and only use the service instance registered at the root of the tree.
```
bennett000: - Break up the sentence `So far our problem is that we are creating two instances of the same services in different levels of the DI tree and that the instance that is created in the lower branch of the tree, is shadowing the one defined at the root level` to `...in different levels of the DI tree. The instance created in the lower branch of the tree is shadowing the one defined at the root level`

handout/modules/lazy-load-di.md
```diff
@@ -0,0 +1,93 @@
+# Lazy Loading and the Dependency Injection Tree
+
+Lazy loaded modules create their own branch on the Dependency Injection (DI) tree. This means that it's possible to have services that belong to a lazy loaded module, that are not accessible by the root module or any other eagerly loaded module of our application.
```
barretodavid: Yes, this is covered in the section "Sharing the Same Dependency Injection Tree"

bennett000:
:shipit:

Issue #767, opened 2016-09-27
Typescript and Typings


barretodavid:
In the section covering how to install type definition files we talked about the typings package to do that but I think there's a new method using directly npm with a syntax like:

```
npm install --save @types/lodash
```

We might need to update the content of that section to reflect the new way to get type definition files.

PR #771, opened 2016-10-04
Update typings section to use @types

watrool:
Connected to https://github.com/rangle/ngCourse2/issues/767

### DIFF
--- a/handout/features/typings.md
+++ b/handout/features/typings.md
@@ -3,9 +3,7 @@
 Astute readers might be wondering what happens when TypeScript programmers need
 to interface with JavaScript modules that have no type information.  TypeScript
 recognizes files labelled `*.d.ts` as _definition_ files.  These files are
-meant to use TypeScript to describe interfaces presented by JavaScript 
+meant to use TypeScript to describe interfaces presented by JavaScript
 libraries.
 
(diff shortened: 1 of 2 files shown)


handout/features/typings.md
```diff
@@ -3,9 +3,7 @@
 Astute readers might be wondering what happens when TypeScript programmers need
 to interface with JavaScript modules that have no type information.  TypeScript
 recognizes files labelled `*.d.ts` as _definition_ files.  These files are
-meant to use TypeScript to describe interfaces presented by JavaScript 
+meant to use TypeScript to describe interfaces presented by JavaScript
 libraries.
 
-There are communities of people dedicated to creating typings for JavaScript
```
bennett000: @watrool are there still some legacy use cases for `typings`?  If so we might want to keep this and be clear that it's deprecated but still exists for legacy reasons.

watrool: @bennett000 Yes, `typings` may still be needed in 1.x versions as they do not support `@types` very well. Please see https://github.com/rangle/ngCourse2/pull/771/commits/df51441f0687639f653feb747d5cd6f018a22b66.

bennett000:
LGTM :shipit:

PR #773, opened 2016-10-04
Added section: Listening to an Element Host

alanthai:
Connected to rangle/ngCourse2#743

Issue #742, opened 2016-09-12
Directives - Host Binding


e-schultz:
As a new Angular 2 developer, I would like to understand why I would make use of host binding in my application

AC:

* Content describing host binding, what it does
* Clear 'real world' example of implementing a directive using host-binding
* Should provide value beyond API docs - https://angular.io/docs/ts/latest/api/core/index/HostBindingMetadata-class.html

Timebox: roughly 0.5 - 1 day depending on familiarity with host binding

PR #774, opened 2016-10-05
Add HostBinding section to book

smithad15:
Connected to #742 

Should be merged after #773 so that the page order can be correct in `SUMMARY.md`


bennett000:
@smithad15 overall, this LGTM.  I was thinking about mentioning something on @alanthai 's PR about being a little bit more explicit about what exactly a host element is, especially in with respect to `@Component` directives.

What do you two think?

smithad15:
I think a bit more clarification on what a host element is would be a good thing to have. It should probably either go in the _HostListener_ section or the _Creating an Attribute Directive_ section. I can add that as a separate PR or merge it into this one depending on your preference.

smithad15:
Host element explanation has been added. Should be GTG barring any requested changes.

bennett000:
nice and straightforward, LGTM :shipit:

Issue #768, opened 2016-09-27
Adding Linting has Changed for Webpack 2


barretodavid:
We are showing this snippet to explain how to add linting to a project with webpack

```js
module: {
  preLoaders: [
    { test: /\.ts$/, loader: 'tslint' }
  ],
  loaders: [
    { test: /\.ts$/, loader: 'ts', exclude: /node_modules/ },
    // ...
  ]
  // ...
}
```

But this has changed for the lastest webpack 2 version to:

```js
module: {
  rules: [
    { test: /\.ts$/, loader: 'tslint', exclude: /node_modules/, enforce: 'pre' },
    { test: /\.ts$/, loader: 'awesome-typescript', exclude: /node_modules/ },
    // ...
  ]
  // ...
}
```

PR #775, opened 2016-10-05
Update webpack config

watrool:
Connected to https://github.com/rangle/ngCourse2/issues/768


bennett000:
@watrool is this all we need to update for the basic webpack info?  Do we need to do anything for Webpack 2?

watrool:
@bennett000 I also updated the vendor file in the config, could you take a look on that? (I am not very sure whether we need to put `es6-shim` there)

And thanks for pointing it out!

bennett000:
This looks like valid webpack 2.0 configuration.  :shipit:

Issue #744, opened 2016-09-12
Real World Examples


e-schultz:
To get better value and understanding of various Angular 2 features, having examples that mimic real-world scenarios and use cases that people have instead of contrived examples that feel forced to just show off the feature would be useful.

Review each section / example in the book, and identify if the example makes sense. If all of the examples in a section seem good / sensible - mark it as done or with a :+1:, if a section contains contrived examples, create a card to improve that example and mark it as :-1: blow. 

This card is to only identify where the work needs to be done, one all sections blow have been marked as :+1: or :-1: - it can be moved to done. 
- [x] Bootstrapping (Troy)
- [x] Directives (Alan Thai)
- [x] Advanced Components (Alan Thai) 
- [x] Dependency Injection (Troy)
- [x] Pipes (Alan Thai)

gillianml:
@bennett000 Thanks for grooming the list above!

PR #789, opened 2016-10-12
Update Dependency Injection Examples

watrool:
Connected to https://github.com/rangle/ngCourse2/issues/744.

### DIFF
--- a/handout/di/angular2/README.md
+++ b/handout/di/angular2/README.md
@@ -5,26 +5,27 @@ simplifies DI even further.  With Angular 2, programmers almost never have to ge
 bogged down with injection details.
 
 Angular 2's DI system is (mostly) controlled through `@NgModule`.  Specifically
-the `providers` array.
+the `providers` and `declarations` array. (`declarations` is where we put components,
+pipes and directives; `providers` is where we put services)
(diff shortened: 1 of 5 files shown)


bennett000:
This certainly does seem more real world.  It's also reasonably agnostic.  I mention this because when we first wrote the DI section @igor-ka really wanted us to use something abstract and removed from frameworks for the _description_ part of the DI chapter.

I'm thinking we might want to keep the text using Hamburger which is completely abstract and _additionally_ provide the chat example that @watrool wrote as an example real world use case.

watrool:
@bennett000 it makes sense to use agnostic example to explain DI that apart from any framework and `Hamburger` is indeed an abstract example. So for this PR, how should we format the two examples: `Hamburger` and `ChatWidget`? Do we need to create a new section "Real world example" under DI?

bennett000:
@watrool the easiest way to split this is probably to use your new examples for everything under `di/angular2` and for the earlier parts keep `Hamburger`

watrool:
@bennett000 I have done the change in 8fe70ec.

bennett000:
:shipit:

PR #791, opened 2016-10-13
I18n chapter

wzaghal:
There's not much material to be found besides [this](https://angular.io/docs/ts/latest/cookbook/i18n.html#!#angular-i18n) which makes it a little difficult for me to figure out how much detail to go into and what details to filter out.

Would love some thoughts on my progress so far.


handout/i18n/README.md
```diff
@@ -0,0 +1,3 @@
+## What is the purpose of i18n?
+
+Internationalization (frequently abbreviated to the numeronym i18n) is the process of adapting software and web applications to multiple languages, allowing their application to be used by multiple language-speaking users. By ensuring that your application supports multiple languages, you can reach a broader audience and ensure a smooth user experience for multilingual users.
```
bennett000: if you're going to mention the abbreviation it might be a good idea to explain that there are 18 characters between i-n in internationalization?

That's a bit nit picky though, but it's interesting.

handout/i18n/import-using-jit.md
```diff
@@ -0,0 +1,69 @@
...
+
+
+declare var System: any;
+function getTranslationsWithSystemJs(file: string) {
+  return System.import(file + '!text'); // relies on text plugin
+}
+
+```
+
+Note that above, we had to create a function `getTranslationsWithSystemJs` and rely
```
bennett000: What if we're not using SystemJS?

bennett000:
huzzah i18n.  This all looks reasonable to me although I'm only just becoming familiar with Angular 2's i18n implementation.  my only major concern is about the build steps in the JIT phase.  I didn't have time to read it thoroughly yet but it doesn't look "Angular-y"... 

I know that's vague but exporting functions (something I love in general) does not seem to be a "normal" pattern I've seen in Angular 2 docs.

handout/i18n/import-using-jit.md
wzaghal: Webpack isn't one of my strong suites. I can look into it though and bug some other folks if I need to.

wzaghal:
@bennett000 Fair point - I'll look through other chapters in the book to see if I can get a better understanding of how I can modify the build steps to make it more like our style, as well as Angular 2's style.

bennett000:
@wzaghal sounds good.  I need to actually play with this myself because I have a few people who I need to explain it to.  Hopefully I can help with the build.

SUMMARY.md
```diff
@@ -288,5 +288,15 @@
...
 ---
 
+* [Internationalization in Angular 2](handout/i18n/README.md)
+    * [What is the process like and how is involved?](handout/i18n/process-and-roles.md)
+    * [Marking text in our templates](handout/i18n/marking-text-in-templates.md)
+    * [Extracting translation text using the Angular CLI](handout/i18n/translation-text-extraction.md)
+    * [How to import the completed translation files](handout/i18n/import-translation-files.md)
+        * [Using the AoT Compiler](handout/i18n/import-using-aot.md)
+        * [Using the JiT Compiler](handout/i18n/import-using-jit.md)
+---
```
barretodavid: The text "Using the JiT compiler" is rendering strangely. I think we need to have a blank line before the section separator.

handout/i18n/import-using-aot.md
```diff
@@ -0,0 +1,21 @@
+# Importing the Translation Files with the AoT Compiler
+
+To Internationalize with the AoT (Ahed of time) compiler, you will have to:
+
+ * pre-build a seperate application package for each language
```
barretodavid: What do you mean by "application package"?

handout/i18n/import-using-jit.md
barretodavid: We should keep an eye on this [SO question](http://stackoverflow.com/questions/40307199/how-do-i-use-the-recently-introduced-i18n-capability-of-angular2-with-an-angular)

barretodavid:
The complete i18n chapter was added to the PR #791 and resolved there

PR #798, opened 2016-10-26
I18n chapter

alanthai:
Piggy back off of rangle/ngCourse2#791


barretodavid:
@alanthai LGTM. Could you squash the commits to have a cleaner git history please?

Issue #805, opened 2016-11-07
Create a chapter for `$http` service


barretodavid:
I think we never even mentioned the `$http` service in the book and that's a recurring question when doing training. We should cover common uses cases and patterns.

PR #809, opened 2016-11-15
added a section for Http

winkerVSbecks:
connects to #805 

cc: @barretodavid @mhazy


handout/http/catching_rejections.md
```diff
@@ -0,0 +1,91 @@
+## Catching Rejections
+
+To catch rejections we use the subscriber's `error` and `complete` callbacks.
+
+```ts
+authenticate(username, password) {
```
barretodavid: For some reason I find it harder to understand the snippet without knowing the name of the class of if you are referring to a component or a service. Would you mind adding the class name to the snippet?

handout/http/catching_rejections.md
```diff
@@ -0,0 +1,91 @@
...
+        );
+      });
+  }
+}
+```
+
+[View Example](http://plnkr.co/edit/XOsH5V?p=preview)
+
+### Cancel a Request
+
```
barretodavid: I think we should help the reader understand why is this important, what would happen if we don't do this (memory leaks, unnecessary use of cpu, corruption of the app state, etc.).

handout/http/catching_rejections.md
```diff
@@ -0,0 +1,91 @@
...
+		    (result) => { this.result = result.artists.items; },
+		    (err) => { this.errorMessage = err.message; },
+        () => { console.log('Completed'); }
+		  );
+
+		request.unsubscribe();
+	}
+}
+```
+
```
barretodavid: As a side note, we should remind the reader to always close any open subscription in the `ngOnDestroy` event to avoid memory leaks.

handout/http/catching_rejections.md
```diff
@@ -0,0 +1,91 @@
...
+  search(term: string) {
+    let tryCount = 0;
+    return this.http.get('https://api.spotify.com/v1/dsds?q=' + term + '&type=artist')
+      .map((response) => response.json())
+      .retry(3);
+  }
+}
+```
+
```
barretodavid: It's worth mentioning that the error callback is never invoked during the retry phase even if an error is thrown in the stream (in this case by the http service when the server responds with a HTTP error number like 500) unless the third retry fails.

handout/http/catching_rejections.md
```diff
@@ -0,0 +1,91 @@
...
+		request.unsubscribe();
+	}
+}
+```
+
+[View Example](http://plnkr.co/edit/9gRSaZ?p=preview)
+
+### Retry
+
```
barretodavid: I think we should again answer the question "why is this important?" as a motivation for the reader

handout/http/catching_rejections.md
```diff
@@ -0,0 +1,91 @@
...
+    .subscribe(
+      authData => this.storeToken(response.id_token),
+      this.logError,
+      () => console.log('Authentication Complete')
+    );
+}
+```
+
+### Catch and Release
+
```
barretodavid: I think we should add a small explanation of why this way to handle errors might be better sometimes than the regular error callback from the subscription. You can find a good reference here: https://github.com/Reactive-Extensions/RxJS/blob/master/doc/gettingstarted/errors.md#catching-errors

handout/http/requests_as_promises.md
```diff
@@ -0,0 +1,40 @@
...
+
+	search() {
+		this.searchService.search(this.searchField.value)
+		  .then((result) => {
+        this.result = result.artists.items;
+      })
+      .catch((error) => console.error(error));
+	}
+}
+```
```
barretodavid: I bet many readers will wonder which approach to follow, should I use the observable or the promise pattern? Some of them, newer to Angular 2, would be inclined to always call `.toPromise` to have something familiar to work with but we should warn them about the downside of doing this: they will be losing the ability to cancel a request and the ability to chain additional rxjs operators to further transform the stream (in the component consuming the service for example).

handout/http/catching_rejections.md
winkerVSbecks: @barretodavid this is a concern for observables in general and not specific to http right? I would capture that in the observable section.

winkerVSbecks:
@barretodavid updated to incorporate your suggestions.

handout/http/README.md
```diff
@@ -0,0 +1,27 @@
...
+    ReactiveFormsModule,
+    FormsModule,
+    HttpModule
+  ],
+  providers: [SearchService],
+  declarations: [MyApp],
+  bootstrap: [MyApp]
+})
+export class MyAppModule {}
+```
```
barretodavid: Would it make sense for you to simplify the snippet a little bit more to highlight the `HttpModule`?

```ts
...
import { HttpModule } from '@angular/http';

@NgModule({
  imports: [
    HttpModule,
    ...
  ],
  ...
})
export class MyAppModule {}
```

handout/http/catching-rejections/catch_and_release.md
```diff
@@ -0,0 +1,51 @@
...
+          return cachedVersion();
+        } else {
+          return Observable.throw(
+            new Error(`${ e.status } ${ e.statusText }`)
+          );
+        }
+      });
+  }
+}
+```
```
barretodavid: Nice addition

handout/http/catching-rejections/cancel_request.md
```diff
@@ -0,0 +1,25 @@
...
+@Component({ ... })
+export class MyApp {
+	...
+
+	search() {
+		const request = this.searchService.search(this.searchField.value)
+		  .subscribe(
+		    (result) => { this.result = result.artists.items; },
+		    (err) => { this.errorMessage = err.message; },
+        () => { console.log('Completed'); }
```
barretodavid: If you see this line directly on the gitbook, there seems to be an indentation problem.

screen shot 2016-11-18 at 4 28 25 pm

handout/http/catching-rejections/cancel_request.md
```diff
@@ -0,0 +1,25 @@
...
+		  );
+
+		request.unsubscribe();
```
barretodavid: I've notice that in the plunker you have this instead:

```ts
setTimeout(() => {
  request.unsubscribe();
  this.errorMessage = 'Request cancelled';
}, 0)
```

Why the `setTimeout` set to 0?

winkerVSbecks: Couldn't come up with any other example of cancelling a request. Should I just leave the `setTimtout` out?

barretodavid: ok now I see what you are trying to do, you want to wait for a vm turn to prove that after the subscription was performed (moved to the event queue) you can unsubscribe from the observable. `setTimeout` makes sense.

handout/http/README.md
bennett000: In some cases we've had commenters complain about the `...` because they think it's destructuring or code. 

I'm not strongly opinionated either way, other than comments potentially making more work.  That said a "busy" example can also make more work in the long run...

winkerVSbecks: how about wrapping the `...` in comments so, `/* ... */`?

winkerVSbecks:
@barretodavid fixed formatting errors and added comments around `...`

barretodavid:
LGTM :shipit:

Issue #734, opened 2016-09-07
DI: Opaque Tokens & Collisions


bennett000:
Add a section to DI about Collisions that covers:
- naming collisions between modules
- Opaque tokens vs strings vs class symbols

PR #812, opened 2016-11-24
Add OpaqueToken page

jamesbs:
Connection to #734

### DIFF
--- a/handout/di/angular2/avoiding_collisions_opaque_token.md
+++ b/handout/di/angular2/avoiding_collisions_opaque_token.md
@@ -1,8 +1,8 @@
 # Avoiding Injection Collisions: OpaqueToken
 
-One of the potential issues with dependency injection is the case where different modules are using the same symbol to represent different entities. If we passed in a string to `@Inject` for example, we could potentially have two different entities that aren't aware of each other with the same identifier to Angular's DI. When it comes time to inject one of the entities, we could be using the wrong value without knowing it. While this behaviour might be difficult to run into or easy to resolve within a project, fixing this problem when we start working with 3rd party modules and services is another issue. This is where Angular's `OpaqueToken` comes into play.
+Since Angular allows the use of tokens as identifiers to its dependency injection system, one of the potential issues is using the same token to represent different entities. If, for example, the string `token` is used to inject an entity, it's possible that something totally unrelated is also using the same string to inject a different entity. Now, when it comes time for Angular to resolve one of them, we could be using a value intended for the other. This behaviour might be difficult to run into or easy to resolve within a small team, but when it comes to multiple teams working separately on the same codebase, or integration of 3rd party modules, the cost of these collisions goes up. This is where Angular's `OpaqueToken` comes into play.
 
-`OpaqueToken`s are unique and immutable values that allow module providers to avoid DI collision issues.
+`OpaqueToken`s are unique and immutable values which allow developers to avoid collisions of dependency injection token ids.
 
 ```
 import { OpaqueToken } from '@angular/core';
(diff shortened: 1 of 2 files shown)


handout/di/angular2/avoiding_collisions_opaque_token.md
```diff
@@ -0,0 +1,86 @@
...
+}
+
+@NgModule({
+  providers: [
+    EmailService,
+  ],
+})
+export class EmailModule { }
+```
+[View Example](https://plnkr.co/edit/esaYGmJK8vOULxHKCgET?p=preview)
```
barretodavid: I've found the code snippets somehow hard to follow, mostly because it's not clear the relationship between `apiConfig` and `emailApiConfig` or `loggerApiConfig`. Also, I think that in a real world scenario, those providers should be defined directly in the feature modules instead of the root module. Take a look at my refactored version of your code [before](https://plnkr.co/edit/vWcZcNAxRFqDkZzL4XEF?p=preview) and [after](https://plnkr.co/edit/56VkwzOVdQsCxRfWVM4Y?p=preview) `OpaqueToken`. What do you think?

handout/di/angular2/avoiding_collisions_opaque_token.md
```diff
@@ -0,0 +1,98 @@
...
+
+_app/email/email.module.ts_
+```typescript
+@NgModule({
+  providers: [ EmailService ],
+})
+export class EmailModule { }
+```
+
+The email service has an associated api that is trying to inject configuration settings identified by the string `'api-config'`.
```
barretodavid: I think we need to explain a little bit more the problem. Below are my proposed changes for this section.

---

The email service requires a configuration object identified by the token `api-service` that will be provided by the DI system. Because we want this module to be used in other applications, and the value of the configuration object depends on the application characteristics, we should provide the value of the token `api-service` in `AppModule` where the `EmailModule` is imported.

_app/app.module.ts_

```ts
import { apiConfig } from './email/index';

@NgModule({
  ...
  providers: [
    { provide: apiConfig, useValue: { apiKey: 'email-config', context: 'registration' } },
  ]
})
export class AppModule {}
```

At the same time, another team is working on a reusable `LoggerModule` that also requires a configuration object. Not surprisingly, they decided to use the same token for their configuration object, `api-config`.

handout/di/angular2/avoiding_collisions_opaque_token.md
```diff
@@ -0,0 +1,98 @@
...
+export class EmailService {
+  constructor(@Inject(apiConfig) public apiConfig: EmailConfig) { }
+}
+
+@NgModule({
+  providers: [ EmailService ],
+})
+export class EmailModule { }
+```
+[View Example](https://plnkr.co/edit/SHfTH9R6JVDwJKnzRFSH?p=preview)
```
barretodavid: I think we should only show the refactoring of the services where the tokens are defined and add some final conclusions. Proposed changes below:

---
_app/email/email.module.ts_

```ts
import { Injectable, Inject, OpaqueToken } from '@angular/core';
import { EmailConfig } from './email.config';

export const apiConfig = new OpaqueToken('api-config');

@Injectable()
export class EmailService {
  constructor(@Inject(apiConfig) public apiConfig: EmailConfig) { }
}
```

_app/logger/logger.service.ts_

```ts
import { Injectable, Inject, OpaqueToken } from '@angular/core';
import { LoggerConfig } from './logger.config';

export const apiConfig = new OpaqueToken('api-config');

@Injectable()
export class LoggerService {
  constructor(@Inject(apiConfig) public apiConfig: LoggerConfig) { }
}
```

[View Example](https://plnkr.co/edit/SHfTH9R6JVDwJKnzRFSH?p=preview)

The rest of the code stays exactly the same as before but now, the collision is avoided. Every service gets the correct configuration object from the root module. With `OpaqueToken`, Angular is now able to differentiate two tokens that uses the same string.

handout/di/angular2/avoiding_collisions_opaque_token.md
```diff
@@ -0,0 +1,98 @@
...
+    { provide: emailApiConfig, useValue: { apiKey: 'email-config', context: 'registration' } },
+    { provide: loggerApiConfig, useValue: { apiKey: 'logger-config' } },
+  ],
+  ...
+})
+export class AppModule { }
+```
+[View Example](https://plnkr.co/edit/QrvjsucT6lF6dnFUb2ag?p=preview)
+
+In this case, module implementation details are leaking out to the parent module. Not only that, those details were obfuscated through the module exports and this can lead to problematic debugging. This is where Angular's `OpaqueToken` comes into play.
```
barretodavid: I think it's going to be easier for the reader if we mention the actual problem after showing the code so he/she will have more context of what we are talking about. My proposed change is to basically reorder some of your current content with some additional details:

---

When we run the application, we can see that we have in fact a collision problem. Both modules are getting the same value for the token `api-config`, in this case `{ apiKey: 'logger-info' }`. When it comes time for the main application to specify those settings, Angular overwrites the first `emailApiConfig` value with the `loggerApiConfig` value, since it was provided last.

In this case, module implementation...

handout/di/angular2/avoiding_collisions_opaque_token.md
```diff
@@ -0,0 +1,98 @@
...
+
+_app/logger/logger.module.ts_
+```typescript
+@NgModule({
+  providers: [ LoggerService ],
+})
+export class LoggerModule { }
+```
+
+`LoggerModule` has an api as well, and it's also using the string `'api-config'` to inject its configuration settings. When it comes time for the main app to specify those settings, Angular overwrites the first `emailApiConfig` value with the `loggerApiConfig` value, since it was provided last.
```
barretodavid: My proposal is to move some of your content after showing the `AppModule`. This is my proposed new paragraph:

---

In an effort to avoid a collision between the two tokens with the same name, we could try to rename the imports as shown below.

handout/di/angular2/avoiding_collisions_opaque_token.md
```diff
@@ -0,0 +1,98 @@
+# Avoiding Injection Collisions: OpaqueToken
+
+Since Angular allows the use of tokens as identifiers to its dependency injection system, one of the potential issues is using the same token to represent different entities. If, for example, the string `'token'` is used to inject an entity, it's possible that something totally unrelated also uses `'token'` to inject a different entity. When it comes time for Angular to resolve one of these entities, it might be resolving the wrong one. This behaviour might happen rarely, or are easy to resolve when it happens within a small team, but when it comes to multiple teams working separately on the same codebase or integration of 3rd party modules, the cost of these collisions goes up.
```
bennett000: > This behaviour might happen rarely, or are easy to resolve when it happens
should read or is easy, also as much as I love Canadian spelling behaviour needs to be behavior because of the style guide.

jamesbs:
Great feedback! Doc has been edited accordingly

jamesbs:
`

barretodavid:
@bennett000 are you ok if we do the merge?

bennett000:
nice I feel more informed, I've never actually tested these collisions. LGTM, :shipit:

Issue #145, opened 2016-03-15
TypeScript Section needs `type` keyword and fat arrow function types


bennett000:

barretodavid:
"Type" Keyword: https://www.typescriptlang.org/docs/handbook/advanced-types.html

turbobeast:
Fat Arrow functions are covered in the ES6 section. Besides the syntax of adding types, they really have no typescript-specific behaviour. Could get into `noImplicitThis`, but that is not specific to arrow functions. 

However, I do think the coverage of functions in typescript is lacking. It doesn't even mention shorthand for instance members in class constructors.

turbobeast:
Nevermind... I see now that it is `fat arrow function TYPES`... I still think that coverage of functions in typescript is lacking.

PR #813, opened 2016-11-24
adds type keyword section -- closes rangle/ngCourse2#145

turbobeast:


barretodavid:
@turbobeast although semicolons are "optional" in Javascript, we have been using it in every documentation and we need to be consistent

barretodavid:
LGTM :shipit:

PR #817, opened 2016-11-26
Component name corrected as well as plunker example.

rajinder-yadav:
To follow ng2 guideline


bennett000:
pedantic but it matters :) :shipit:

PR #823, opened 2016-11-26
Expanded example, added missing quotes.

rajinder-yadav:


handout/directives/ng_style_directive.md
```diff
@@ -24,3 +24,27 @@ export class StyleExampleComponent {
...
+    <p style="padding: 1rem"
+        [ngStyle]="alertStyles">
+        <ng-content></ng-content>
+    </p>
+  `
+})
+export class StyleExampleComponent {
+  borderStyle: string = '1px solid black';
+
+  // Must be JSON valid.
```
mhazy: I'd remove this comment, it's a plain object. The property values can also be functions that return values.

rajinder-yadav: Thanks

rajinder-yadav: Let me think of how to reword this comment!

handout/directives/ng_style_directive.md
```diff
@@ -40,7 +40,7 @@ We can remove the style properties out of the template into the component as a p
 export class StyleExampleComponent {
   borderStyle: string = '1px solid black';
 
-  // Must be JSON valid.
+  // Properties must be quoted.
```
mhazy: It just has to be enumerable object. Quotes are only required if it's a plain 'ol object and the CSS property contains a dash, which would not be valid, e.g. `border-radius`, in which case you may use camel case property values, e.g. `borderRadius`.

Here's some different ways to use ngStyle (e.g. using a class with properties defined programatically)

https://plnkr.co/edit/bDTs3s?p=preview

bennett000:
LGTM :shipit:

PR #825, opened 2016-11-26
Paragraph separated wit own heading.

rajinder-yadav:


bennett000:
This is an interesting one.  In writing we use `*ngFor` but the class that makes the directive is `NgFor`.  I don't know what the correct style is.  I'm inclined to agree with @rajinder-yadav 's PR but I'd like input from someone else, maybe @barretodavid or @mhazy

barretodavid:
I agree with @rajinder-yadav, I think we should use the name of the class `NgFor` instead of `*ngFor`.

mhazy:
Yeah, I agree, `NgFor` when talking about the directive as a feature. It's how it's referenced in the docs as well, https://angular.io/docs/ts/latest/api/common/index/NgFor-directive.html

bennett000:
:shipit:

Issue #519, opened 2016-06-27
Update examples/exercises to be consistent with style guide


e-schultz:
When doing the initial content/examples - was several people working on it and no 'style guide' yet.

We have a mix of naming conventions, how things are structured, etc. 

Should try to update things in terms of

* inline code examples
* plunkrs
* exercise folder
* example folder

To be consistent, and closer to the https://angular.io/styleguide

alejochang:
- [x] add linting 
- [] exercise folder
- [] example folder
- [] plunkers
- [] inline code examples

alejochang:
Adding linting to perform an initial clean up. I'll use it to help me to update the example/exercises incrementally but it'll be added at the end with the last PR.

alejochang:
I'm doing a pull request by section, which covers code example / plunkers and inline code updates

PR #826, opened 2016-11-27
"Creating Components" section - angular style guide update

alejochang:
Updated "Creating Components" section code example / plunker to be closer to the angular 2 style guide. #519


bennett000:
:shipit:

PR #830, opened 2016-11-28
"Using Two-Way Data Binding" section - angular style guide update

alejochang:
Updated code example / plunker to be closer to the angular 2 style guide. #519


examples/components/components-two-way-binding/app/counter.component.ts
```diff
@@ -1,22 +1,17 @@
...
-      </p>
-
-    </div>
-  `
+  selector: 'rio-counter',
+  templateUrl: 'app/counter.component.html'
 })
-export class Counter {
-  @Input() count: number = 0;
-  @Output() countChange: EventEmitter<number> = new EventEmitter<number>();
```
bennett000: Like on the other PR, is it an Angular style guide practice to initialize in the constructor?

alejochang: Changed to use shortcut version.

Issue #640, opened 2016-07-29
Self Closing Tags


imsingh:
Angular 2 is not currently supporting self closing tags. But on 67th page of book, it shows a HTML Snippet with self closing tags.

bennett000:
We will be fixing this and adding an explicit note at the start of our templates section about this.

watrool:
@bennett000 @imsingh Where are these tags? We changed the book a lot and current 67th page does not have any code snippet.

PR #838, opened 2016-12-13
Remove instances of self closing tags in handouts and plunker

jamesbs:
Connects to #640


bennett000:
Once this is rebased it can :shipit:

PR #846, opened 2016-12-16
Update modules section with angular 2 style guide

jamesbs:
Connects to #519


bennett000:
so we went from `rio` prefix to `app` prefix.  I think both made sense but `app` is more generic so that seems to make more sense.

I know why there were so many "touches" for this file.  During today's meeting we'll try and strategize on how to reduce PR size.

:shipit:

jamesbs:
Yeah I think `app` works better for the book due to it being generic. I know in training engagements a common question is what the `rio` is for in examples. I've also being trying to match the presets from `ng-cli` when possible to reduce potential friction.

PR #850, opened 2016-12-19
update pipes section and examples to match style guide

jamesbs:
Connects to #519

Main trade off is the switch the general `AppComponent` naming which is less descriptive than `ProductPrice` for example, but I don't think it affects how people follow the examples


bennett000:
Bigger diff than I would have expected.  Thanks.  :shipit:

PR #872, opened 2017-01-13
Removed in-place explanation and added reference to arrow functions s…

### LINKED ISSUE #871: Add reference to arrow function in the current book and omit in-place explanation
https://angular-2-training-book.rangle.io/handout/features/es6_modules.html

The above section has an in-place explanation about arrow functions instead add a reference to the arrow functions section in the current book.

ashwin-sureshkumar:
Added reference for arrow functions in the current book.  resolves #871

### DIFF
--- a/handout/features/es6_modules.md
+++ b/handout/features/es6_modules.md
@@ -13,13 +13,7 @@ export const pi = 3.141592;
 export const circumference = diameter => diameter * pi;
 ```
 
-The code above uses the _Arrow_ function for `circumference`, which was introduced in ES6, and is a shortform for the following.
-
-```js
-export function circumference(diameter) {
-  return diameter * pi;
-}
-```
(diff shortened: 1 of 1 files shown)


handout/features/es6_modules.md
```diff
@@ -13,13 +13,7 @@ export const pi = 3.141592;
...
 ```
 
-The code above uses the _Arrow_ function for `circumference`, which was introduced in ES6, and is a shortform for the following.
-
-```js
-export function circumference(diameter) {
-  return diameter * pi;
-}
-```
+The code above uses the _Arrow_ function for `circumference`. Read more about arrow functions [here](https://angular-2-training-book.rangle.io/handout/features/arrow_functions.html)
```
barretodavid: This is my recommendation for the paragraph:

"The code above uses an _arrow_ function for `circumference` which was introduced in ES6. To learn more about, visit the [Arrow Function](./arrow_functions.md) section in the book."

Why?

- We should make obvious to the user that it's an internal link, not external.
- We should use relative links pointing to the markdown file so we can use the book in a dev environment as well.

PR #880, opened 2017-01-13
Update rest and spread sections according to specification

### LINKED ISSUE #879: Update Rest and Spread operator section to reflect spec
https://angular-2-training-book.rangle.io/handout/features/spread_and_rest.html

In the above section, the names and argument reference are bit unclear and not completely according to spec of ECMAScript

ashwin-sureshkumar:
Resolves #879

### DIFF
--- a/handout/features/spread_and_rest.md
+++ b/handout/features/spread_and_rest.md
@@ -35,7 +35,7 @@ let mapABCD = { ...mapABC, d: 7};  // { a: 5, b: 6, c: 3, d: 7 }
 
 ## Rest parameter
 
-Rest parameters share the ellipsis like syntax of spread operator but are used
+Rest parameters share the ellipsis like syntax of spread syntax but are used
 for a different purpose. Rest parameters are used to access indefinite number
 of arguments passed to a function. For example:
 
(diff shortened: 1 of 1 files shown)


handout/features/spread_and_rest.md
```diff
@@ -33,10 +33,10 @@ let mapABC  = { a: 5, b: 6, c: 3};
 let mapABCD = { ...mapABC, d: 7};  // { a: 5, b: 6, c: 3, d: 7 }
 ```
 
-## Operator Rest
+## Rest parameter
 
-Rest arguments share the ellipsis like syntax of rest operators but are used
-for a different purpose. Rest arguments are used to access a variable number
+Rest parameters share the ellipsis like syntax of spread operator but are used
```
barretodavid: I thought we were trying to avoid the phrase "spread operator" don't we?

handout/features/spread_and_rest.md
```diff
@@ -61,9 +61,8 @@ Technically JavaScript already had an `arguments` variable set on each function
 (except for arrow functions), however `arguments` has a lot of issues, one of
 which is the fact that it is not technically an array.
 
-Rest arguments are in fact arrays. The other important difference is that
-rest arguments only include arguments not specifically named in a function
-like so:
+Rest parameters are in fact arrays which provides access to methods like `map, filter, reduce and more`. The other important difference is that
+rest parameters only include arguments not specifically named in a function like so:
```
barretodavid: It should be in one "line" (one paragraph)

handout/features/spread_and_rest.md
```diff
@@ -33,10 +33,10 @@ let mapABC  = { a: 5, b: 6, c: 3};
 let mapABCD = { ...mapABC, d: 7};  // { a: 5, b: 6, c: 3, d: 7 }
 ```
 
-## Operator Rest
+## Rest parameter
 
-Rest arguments share the ellipsis like syntax of rest operators but are used
-for a different purpose. Rest arguments are used to access a variable number
+Rest parameters share the ellipsis like syntax of spread syntax but are used
+for a different purpose. Rest parameters are used to access indefinite number
```
barretodavid: Why two different lines?

Also, proposed change: 

"Rest parameters look very similar to the "spread syntax" but are used for a different purpose. Rest parameters are used to access indefinite number of arguments passed to a function. For example:"

handout/features/spread_and_rest.md
```diff
@@ -1,14 +1,14 @@
-# Operators Spread and Rest
+# Spread Syntax (Spread Element) and Rest parameters
 
-A Spread operator allows in-place expansion of an expression for the following cases:
+A Spread syntax allows in-place expansion of an expression for the following cases:
```
bennett000: I guess `...` is [not explicitly an operator](http://www.ecma-international.org/ecma-262/6.0/ "ES 6.0 Spec") although I don't feel like calling it a syntax is quite correct either.  The official term "Punctuator" also seems awkward @barretodavid any thoughts on this?

barretodavid: @ashwin-sureshkumar showed me the [MDN reference page](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Operators/Spread_operator) and they claim that to be a syntax so I agreed to change the name. Interestingly, the url for that page has at the end "Spread_operator". Maybe they left the url for SEO reasons.

barretodavid: Also, it's weird that at the end they are using the term "Rest operator" when in another section they are referring to "Rest parameters". I don't know what to believe anymore...

barretodavid: To add more confusion to the mix. If we refer always to "Rest parameters" then what term should we use to refer to this?

```js
[a, b, ...rest] = [1, 2, 3, 4, 5];
```

PR #884, opened 2017-01-14
New Page for HTTP section on Using switchMap

### LINKED ISSUE #855: Handling race condition with `switchMap` for the `$http` section
At the end of the section [Combining Streams with `flatMap`](https://angular-2-training-book.rangle.io/handout/http/search_with_flatmap.html) one of our readers made a good point:

> I don't think flatMap is an ideal usage for incremental search refer this http://shuheikagawa.com/blog/2016/05/01/incremental-search-with-rxjs/

When the server takes longer that the debounce time to response, there's a change that we get the response in the wrong order and thus showing the wrong result. `switchMap` will help us avoid that problem.

We should add a new page stating this issue and why we should use `switchMap` instead of `flatMap` when there's a risk of race conditions.

adrianmcli:
Fixes #855 

This is a first draft, I am looking for people to review/edit it. Any and all comments and feedback are welcome.

### Summary of Problem

There was a comment indicating that the page on flatMap proposed an inappropriate/insufficient solution to an autocomplete input/incremental search implementation. I agree with this and have been tasked to create a new page to explain switchMap and how it will help to make the incremental search component more robust.

Please see the original issue #855 for more detail.


handout/http/search_with_switchmap.md
```diff
@@ -0,0 +1,91 @@
...
+// A1: Request for `ABC`
+// A2: Response for `ABC`
+// B1: Request for `ABCX`
+// B2: Response for `ABCX`
+
+--A1----------A2-->
+------B1--B2------>
+```
+
+You can see that A2 arrives after B2 even though the A1 request began first. So how do we get around this problem? We can use `switchMap`.
```
barretodavid: Proposed change:

You can see that A2 arrives after B2 even though the A1 request began first. This will end up showing the wrong results to the user. "If the last input in the search was `ABCX` why am I seeing the results for `ABC`?" the user might think. To get around this problem we need to replace `flatMap` with `switchMap`.

handout/http/search_with_switchmap.md
```diff
@@ -0,0 +1,91 @@
...
+
+## What is `switchMap`?
+
+SwitchMap is very similar to `flatMap`, but with a very important distinction. Any events to be merged into the trunk stream are ignored if a new event comes in. Here is a marble diagram showing the behavior of `switchMap`:
```
barretodavid: `swtichMap` is very similar to `flatMap`...

handout/http/search_with_switchmap.md
```diff
@@ -0,0 +1,91 @@
...
+
+![SwitchMap created by ReactiveX licensed under CC-3 (http://reactivex.io/documentation/operators/flatmap.html)](../images/switch-map.png)
+
+Just like `flatMap`, the red marble gets replaced with a red diamond and a subsequent red square. The interaction between the green and blue marble events are more interesting. Note that the green marble gets mapped to a green diamond immediately. And if enough time had passed, a green square would be pushed into the trunk stream but we do not see that here.
+
+Before the green square event is able to happen, a blue marble comes through and gets mapped to a blue diamond. What happened is that the green square is now ignored and do not get merged back into the trunk stream. The behavior of `switchMap` can be likened to a `flatMap` that "switches" to the more immediate incoming event and ignores all previously created event streams.
```
barretodavid: I would add an extra paragraph for clarity:

In short, every time an event comes down the stream, `flatMap` will subscribe to a new observable without unsubscribing to any other observable created by a previous event. `switchMap` on the other hand will automatically unsubscribe to any previous observable when a new event comes down the stream.

handout/http/search_with_switchmap.md
```diff
@@ -0,0 +1,91 @@
...
+
+In our case, because the blue marble event happened very quickly after the green marble, we "switched" over to focus on dealing with the blue marble instead. This behavior is what will prevent the problem we described above.
+
+If we apply `switchMap` to the above example, the response for `ABC` would be ignored and the suggestions for `ABCX` would remain.
```
barretodavid: In general I found a little difficult to understand the explanation referring to just colors without any concrete reference to the problem at hand. It would be useful in my opinion to explain that the "diamonds" mark the creation (and subscription) of an inner observable and that the "square" represents values emitted from that inner observable and passed down the stream.

handout/http/search_with_switchmap.md
```diff
@@ -0,0 +1,91 @@
...
+		this.searchField.valueChanges
+		  .debounceTime(400)
+			.switchMap(term => this.searchService.search(term))
+			.subscribe((result) => {
+				this.result = result.artists.items
+			});
+	}
+}
+```
+[View Example](http://plnkr.co/edit/FYLTcx?p=preview)
```
barretodavid: We need to add a summary at the end. Why is this approach better than the one with `flatMap`?

adrianmcli:
@barretodavid I've made changes with regards to your request for changes.

PR #885, opened 2017-01-16
Issue 843 angular2 to angular

ShaunLloyd:
### Update all references of `Angular 2` to `Angluar`

- In cases where the file just refers to `Angular 2`, all instances are corrected to `Angular`

- In cases where `Angular 1.x` and `Angular >=2` are referenced, Instances are corrected to `Angular` or `Angular >= 2` depending on the context.


examples/ng-upgrade/README.md
```diff
@@ -1,8 +1,8 @@
 # ng-upgrade example
 
-This example is a stand alone Angular 1.x / Angular 2 hybrid application and
+This example is a stand alone Angular 1.x / Angular hybrid application and
```
barretodavid: Same here, when comparing two versions of Angular I would use "Angular 1 / Angular 2" like "Chrome 54 / Chrome 55" like Google is doing for their browser.

handout/change-detection/angular_1_vs_angular_2.md
```diff
@@ -1,4 +1,4 @@
-# Change Detection Strategies in Angular 1 vs Angular 2
+# Change Detection Strategies in Angular 1.x vs Angular >= 2
```
barretodavid: Again, leave it as is

handout/change-detection/angular_1_vs_angular_2.md
```diff
@@ -8,8 +8,8 @@ Because of the nature of two-way data binding, in Angular 1 there was no guarant
 10 $digest() iterations reached. Aborting!
 ```
 
-In Angular 2, changes are guaranteed to propagate unidirectionally. The change detector will **traverse each node only once**, always starting from the root. That means that a parent component is always checked before its children components.
+In Angular, changes are guaranteed to propagate unidirectionally. The change detector will **traverse each node only once**, always starting from the root. That means that a parent component is always checked before its children components.
 
-_Tree traversing in Angular 1 vs Angular 2_
+_Tree traversing in Angular 1.x vs Angular >= 2_
```
barretodavid: idem

handout/components/creating_components.md
```diff
@@ -1,6 +1,6 @@
 # Creating Components
 
-Components in Angular 2 build upon the lessons learned from Angular 1.5. We define a component's application logic inside a class. To this we attach `@Component`, a TypeScript `decorator`, which allows you to modify a class or function definition and adds metadata to properties and function arguments.
```
barretodavid: Again, in this case use Angular 2 and Angular 1.5 because we are comparing two specific versions.

Also, I think the premise of this paragraph is wrong. I'm pretty sure that the `.component` function of Angular 1.5 is a backport from Angular 2 to ease the migration. It's worth investigating that and creating a separate PR if the paragraph needs to be updated.

handout/di/README.md
```diff
@@ -1,5 +1,5 @@
-# Angular 2 Dependency Injection
+# Angular Dependency Injection
 
-Dependency Injection (DI) was a core feature in Angular 1.x, and that has not changed in Angular 2.  DI is a programming concept that predates Angular. The purpose of DI is to simplify dependency management in software components. By reducing the amount of information a component needs to know about its dependencies, unit testing can be made easier and code is more likely to be flexible.
+Dependency Injection (DI) was a core feature in Angular 1.x, and that has not changed in Angular >= 2.  DI is a programming concept that predates Angular. The purpose of DI is to simplify dependency management in software components. By reducing the amount of information a component needs to know about its dependencies, unit testing can be made easier and code is more likely to be flexible.
```
barretodavid: For this section keep with "Angular 2" because of the comparison with Angular 1

SUMMARY.md
```diff
@@ -118,7 +118,7 @@
 ---
 
 * [Change Detection](handout/change-detection/README.md)
-   * [Change Detection Strategies in Angular 1 vs Angular 2](handout/change-detection/angular_1_vs_angular_2.md)
+   * [Change Detection Strategies in Angular 1 vs Angular >= 2](handout/change-detection/angular_1_vs_angular_2.md)
```
ShaunLloyd: thanks David, I shall change that now

bennett000:
:shipit:

PR #886, opened 2017-01-16
remove reimport of core root module in lazy module, closes #883

### LINKED ISSUE #883: Sharing the same dependency injection tree - we shouldn't need to reimport SharedModule
In the Modules section: [Sharing the Same Dependency Injection Tree](https://angular-2-training-book.rangle.io/handout/modules/shared-di-tree.html#), we reimport `SharedModule` into our lazily loaded module which isn't needed if we define it as a core module in our application root with `forRoot`. [This can apparently cause issues as well.](https://angular.io/docs/ts/latest/guide/ngmodule.html#!#prevent-reimport-of-the-_coremodule_)

Will put up a PR to update this page and the plunker as well.

housseindjirdeh:
Removes reimport of `SharedModule` in lazy loaded module (defined with `forRoot` in application root module)

### DIFF
--- a/handout/modules/shared-di-tree.md
+++ b/handout/modules/shared-di-tree.md
@@ -39,8 +39,60 @@ import { SharedModule } from './shared/shared.module';
 export class AppModule {}
 ```
 
-Since we import the module with `forRoot` in our root application module, we **do not** need to import it in `LazyModule`. Reimporting core root modules defined with `forRoot` in lazily loaded modules can cause issues in your application.
+We should **only** call `forRoot` in the root application module and no where else. This ensures that only a single instance of your service exists at the root level. Calling `forRoot` in another module can register the service again in a different level of the DI tree. 
+
+Since `SharedModule` only consists of a service that Angular registers in the root app injector, we do not need to import it in `LazyModule`. This is because the lazy loaded module will already have access to services defined at the root level.
 
 [View Example](https://plnkr.co/edit/dPpc40plyI8iu8ogrf3e?p=preview)
 
 This time, whenever we change the value of the `counter` property, this value is shared between the `EagerComponent` and the `LazyComponent` proving that we are using the same instance of the `CounterService`.
+
+However it is very likely that we may have a component, pipe or directive defined in `SharedModule` that we'll need in another module. Take the following for example.
+
+_app/shared/shared.module.ts_
+
+```ts
+import { NgModule, ModuleWithProviders } from '@angular/core';
+import { CounterService } from './counter.service';
+
+import { HighlightDirective } from './highlight.directive';
+
+@NgModule({
+  declarations: [HighlightDirective],
+  exports: [ HighlightDirective ]
+})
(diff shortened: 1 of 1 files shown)


handout/modules/shared-di-tree.md
```diff
@@ -39,24 +39,8 @@ import { SharedModule } from './shared/shared.module';
 export class AppModule {}
 ```
 
-In contrast, when import the same module in our `LazyModule` we will not call the `forRoot` method because we don't want to register the service again in a different level of the DI tree, so the declaration of the `LazyModule` doesn't change.
+Since we import the module with `forRoot` in our root application module, we **do not** need to import it in `LazyModule`. Reimporting core root modules defined with `forRoot` in lazily loaded modules can cause issues in your application.
```
barretodavid: We need to explain the issues or point to a section where the issue is explained

barretodavid:
@housseindjirdeh I think that the problem with this section is that we are not defining a component to share between the root module and the lazy loaded to better explain the problem. Right now importing the `SharedModule` in the `LazyModule` is not needed because the child injector created by the lazy loaded module will still have access to the `CounterService` defined in the root injector. But we need to explain a more general problem. What if we have also a component (or a pipe or a directive) define in the `SharedModule` that we need to use in the `RootModule` and `LazyModule`? That's the real question for this section.

handout/modules/shared-di-tree.md
housseindjirdeh: Thanks, will do that 👍

housseindjirdeh:
Thanks for this @barretodavid, and you're right it would make more sense to explain how we can import such a module. Will write up another section to explain what to do in this scenario by tomorrow.

handout/modules/shared-di-tree.md
bennett000: What kind of issues?  Multiple states or just duplicated downloads?  (Sorry not trolling, genuinely curious)

housseindjirdeh: Yep so firstly the docs mention that calling `forRoot` in another module, particularly lazy loaded modules, can trigger a run time error. However, if `SharedModule` provides a service to a lazy loaded module (through `providers` for example), the lazy loaded module would make it's own copy of the service.

Could definitely be more clear when I cover it in this section.

housseindjirdeh:
Updated the section to explain things in a little more detail. I first explain how you don't have to reimport `SharedModule` with only a core service since it's defined at the root level. I then introduce a `HighlightDirective` to explain why you may still need to import the module in your lazy loaded module.

Let me know if you think I should change anything!

bennett000:
@barretodavid if you have capacity can you see if your AC is met?

barretodavid:
Good job explaining the problem @housseindjirdeh. I know a lot of people outside of Rangle have a hard time understanding how the injector tree works with lazy loading. They might find this section helpful.

housseindjirdeh:
Thank you David 😊

PR #888, opened 2017-01-16
Minor fixes in routing section, closes #887

### LINKED ISSUE #887: Minor updates to routing section
Few comments on routing section for the following issues:

1. Add an explanation of `pathMatch: 'full'` in [redirect section](https://angular-2-training-book.rangle.io/handout/routing/redirects.html)
2. Routing defined as export in same file in [Configuring Routes section](https://angular-2-training-book.rangle.io/handout/routing/config.html)
3. Square brackets (`[routerLink]`) when using a literal string which is not needed (in a few different sections)

Will have a PR up soon to fix this

housseindjirdeh:


handout/routing/redirects.md
```diff
@@ -11,6 +11,7 @@ export const routes: Routes = [
 ];
 ```
 
-This tells the router to redirect to component-one when matching the empty path ('').
+The `pathMatch` property, which is required, tells the router how it should match the URL provided in order to redirect to the specified route. 
+Since `pathMatch: full` is provided, the router will redirect to `component-one` if the entire URL matches the empty path ('').
```
barretodavid: Are these two sentences different paragraphs or the same one? If two different paragraphs we need to add a blank line. If the same paragraph you should remove the break line

handout/routing/config.md
```diff
@@ -49,6 +51,13 @@ const routes: Routes = [
 ];
 
 export const routing = RouterModule.forRoot(routes);
+```
+
+_app/index.ts_
+
+```javascript
```
barretodavid: btw, the correct syntax highlight to use is `ts` not `javascript` in this case

handout/routing/redirects.md
```diff
@@ -11,6 +11,7 @@ export const routes: Routes = [
 ];
 ```
 
-This tells the router to redirect to component-one when matching the empty path ('').
+The `pathMatch` property, which is required, tells the router how it should match the URL provided in order to redirect to the specified route.
```
barretodavid: "...which is required **for redirects**..."

handout/routing/config.md
```diff
@@ -49,6 +51,13 @@ const routes: Routes = [
 ];
 
 export const routing = RouterModule.forRoot(routes);
+```
+
+_app/index.ts_
```
housseindjirdeh: Yep I was about to do that but the later plunker examples use `index.ts`. I'll update each of the plunkers to use `app.module.ts` instead.

handout/routing/config.md
housseindjirdeh: Oh thank you I wasn't sure if there syntax highlighting for TS

housseindjirdeh:
Updated based on review, also modified each plunker in section to follow styleguide of having `app.module.ts` as the root not `index.ts`. Child routes accessing parent routes plunker wasn't working as well which I fixed.

handout/routing/routeroutlet.md
```diff
@@ -22,6 +22,6 @@ export class AppComponent {}
 
 In the above example, the component corresponding to the route specified will be placed after the `<router-outlet></router-outlet>` element when the link is clicked.
 
-[View Example](https://plnkr.co/edit/odsZq1pL43HtWcUr0taJ?p=preview)
+[View Example](https://plnkr.co/edit/OHfytJquXKm8jvSe2T9Y?p=preview)
```
bennett000: We used to have a link on here that would open the example in a new window and actually allow the routes to change.  Can we get that working again?  I was a little surprised by the `[routerLink]` -> `routerLink` change and I wanted to see it in action.

housseindjirdeh: Ah you're right `plunker`'s previewer doesn't show route changes in the URL. Looked into just referencing a link that would just generate the full screen view but doesn't seem to work for some reason (http://run.plnkr.co/plunks/odsZq1pL43HtWcUr0taJ). 

There's a note right after this link that mentions you have to open it in full screen mode in order to see the route changes and that works just fine (the little blue icon top right underneath your username when you open the example). Do you think this is alright? Other examples in this chapter mention the same thing and unfortunately I don't think there's another way we can do this with `plunker`.

housseindjirdeh: Ah you're right plunker's previewer doesn't show route changes in the URL. Looked into just referencing a link that would just generate the full screen view but doesn't seem to work for some reason (http://run.plnkr.co/plunks/odsZq1pL43HtWcUr0taJ).

There's a note right after this link that mentions you have to open it in full screen mode in order to see the route changes and that works just fine (the little blue icon top right underneath your username when you open the example). Do you think this is alright? Other examples in this chapter mention the same thing and unfortunately I don't think there's another way we can do this with plunker.

PR #917, opened 2017-05-25
[architect's guide] adds a chapter on 10 Ways to Misuse Angular

winkerVSbecks:
Expanded on the outline created by @rpdasilva. I tried to link to existing content where possible.
And I don't know enough about forms to expand **point 7**.

cc: @rangle/angular-content-maintainers

### DIFF
--- a/handout/architect/10_ways_to_misuse_angular.md
+++ b/handout/architect/10_ways_to_misuse_angular.md
@@ -1,0 +1,439 @@
+# Ten Ways to Misuse Angular
+
+## 1. Ignoring a component’s host element
+When creating Angular components you should use the host element as a wrapper instead of creating your own wrapper elements.
+
+The card component below renders a heading and a paragraph. In order to control their layout we need a wrapper element. You might be tempted to wrap them in a `<div>`, but that is not required. Angular will render this with a `<my-card>` wrapper element. This is called a host element.
+
+```ts
+import { Component } from '@angular/core';
+
+@Component({
+  selector: 'my-card',
+  template: `
+  <h1 class="card-title">{{ title }}</h1>
+  <p class="card-body">
+    <ng-content></ng-content>
+  </p>
+  `
+})
+export class CardComponent { title = 'Title of card'; }
+```
+
+We can style the host element by targeting it with the `:host` selector in the component's CSS file. Alternatively you can attach CSS classes to it using [host metadata](https://angular.io/docs/ts/latest/cookbook/ts-to-js.html#!#host-metadata).
+
+```ts
+@Component({
+  selector: 'my-card',
+  host: { class: 'db ma0 lh-title' },
+  ...
+})
+export class CardComponent { title = 'Title of card'; }
+```
+
+We can use host metadata to bind to all properties and events of the host element. Alternatively, you can use the `@HostBinding` and `@HostListener` decorators.
+
+### Metadata vs decorators
+Host metadata and `@Host*` decorators can generally be used interchangeably however, there are some subtle differences. Consider the following scenario:
+
+```ts
+@Component({
+  selector: 'my-foo',
+  host: { 'class': 'navy' },
+})
+export class FooComponent {}
+
+@Component({
+  selector: 'my-bar',
+})
+export class BarComponent {
+  @HostBinding('class') className = 'navy';
+}
+```
+
+and somewhere later in our app you use them as:
+
+```html
+<my-foo class="underline"></my-foo>
+<my-bar class="underline"></my-bar>
+```
+
+`<my-foo>` will render with `class="underline navy"` however, `<my-bar>` will render with `class="navy"`. This is because `@HostBinding` only supports binding of values. Whereas, host metadata supports both binding and static values. In this case we are using a static value. Therefore, it merges what we have defined in the metadata with what is defined on the component. If we had bound class to a component property then they would behave the same:
+
+```ts
+@Component({
+  selector: 'my-foo',
+  host: { '[class]': 'className' },
+})
+export class FooComponent {
+  className = 'navy';
+}
+```
+
+
+
+## 2. Avoiding observables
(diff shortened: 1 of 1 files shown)


handout/architect/10_ways_to_misuse_angular.md
```diff
@@ -0,0 +1,439 @@
...
+}
+```
+
+
+
+## 2. Avoiding observables
+When working with Angular you should be using observables as much as possible. Angular exposes HTTP requests, form events, route params, etc. as observables. Therefore, by using observables we can build apps that have a consistent architectural pattern.
+
+### Leverage Observable Operators
+We can leverage observable operators to compute data based on application state. This allows us to reduce the amount of state we need to maintain. Additionally these computations can expressed as pure functions which makes testing easier. These pure play a similar to role to [selectors](https://github.com/reactjs/reselect) and can be memoized.
```
rpdasilva: s/these computations can expressed/these computations can be expressed/
s/These pure play/These pure functions play/

handout/architect/10_ways_to_misuse_angular.md
```diff
@@ -0,0 +1,439 @@
...
+
+This pattern can be adopted regardless of whether you are using ngrx, angular-redux or even just services to manage state. When woking with Angular services you can replicate a _store_ style API using Behavior Subjects. In the example below we have a `TodoStore` service. It exposes the entire store as the `state$` observable. To display a list of filtered todos we can simply `map` onto the state$` observable instead of maintaining multiple lists.
```
rpdasilva: s/When woking with Angular/When working with Angular/

handout/architect/10_ways_to_misuse_angular.md
```diff
@@ -0,0 +1,439 @@
...
+      todos.filter(todo => this.filterTodoByType(todo, filter));
+  }
+}
+```
+
+Observables also allow us to merge multiple streams to implement complex workflows such as a [debounced search which ignores stale requests](../http/search_with_flatmap.md) in a declarative manner.
+
+
+### Subscribing to Data
+To access a data property use operators such as `map` or `pluck` to derive separate streams. This allows you to utilize the `async` pipe in the templates and removes the need for managing subscriptions in the component class.
```
rpdasilva: Maybe it's worth mentioning that like Promises, observables beget observables, and that you can "cache" a stream for reuse. Any operators or transformations that occurred in that base observable get passed on to any subsequent "child" or "branch" observables

handout/architect/10_ways_to_misuse_angular.md
```diff
@@ -0,0 +1,439 @@
...
+
+When accessing multiple properties subscribing to the base observable will likely be a better option.
```
rpdasilva: Is it worth providing an example of what attempting to use the async pipe with multiple properties could make your templates look like?

handout/architect/10_ways_to_misuse_angular.md
```diff
@@ -0,0 +1,439 @@
...
+
+For more detailed examples see [Lazy Loading and the Dependency Injection Tree](../modules/lazy-load-di.md), [Creating a Feature Module](../modules/feature-modules.md) and [Sharing the Same Dependency Injection Tree](../modules/shared-di-tree.md) sections.
+
+If you would like to provide a service as an application-level singleton use the `forRoot()` function. For providing services to feature modules you should always use `forChild()`.
+
+
+
+
+## 5. Confusing the Source of Truth Between Router and Application State
+The angular router allows us to define a client side routing structure and manages transitions between those routes. It does not however, manage or mutate your application state. You are responsible for updating the application state in response to navigation events.
```
rpdasilva: You certainly hit the main points of how you _should_ use the router, but the problem it's solving is still left a little unclear.

May be worth mentioning that it's "tempting" or "common" to find code trying to manually manage state as well as router events as two separate things rather than treating state changes such as "selecting an item" as a route transition to being with (i.e. triggered by `router.navigate()` vs. dispatching an action to update the store)

handout/architect/10_ways_to_misuse_angular.md
```diff
@@ -0,0 +1,439 @@
...
+    - Forms are generated based on some sort of schema
+    - Primary concern is the display model of the form rather than the data model
+    - Making composable model-driven form components requires more work
+        - Passing `FormControl` references or implementing the `ControlValueAccessor` interface
+
+
+
+
+## 8. Using Redux Directly Inside of Components
+We can divide components into two types: [Presentational and Container Components](https://medium.com/@dan_abramov/smart-and-dumb-components-7ca2f9a7c7d0). Where Presentational components are responsible for how things look. They receive data through `@Input` and handle events by invoking callbacks or by dispatching events through `@Output`. Container components are responsible to defining how things work i.e., fetching data and updating state.
```
rpdasilva: s/responsible to defining/responsible for defining/

handout/architect/10_ways_to_misuse_angular.md
```diff
@@ -0,0 +1,439 @@
...
+  </Block>
+</Card>
+```
+
+For more on this subject see the [Flattening Deep Hierarchies of Components](http://blog.rangle.io/flattening-deep-hierarchies-of-components) post on the Rangle.io blog.
+
+
+
+
+## 7. Using model-driven forms straight away
```
rpdasilva: @winkerVSbecks I think just focusing on the main points here would be sufficient. The outline for this part may have been a little... over-thought. Let me know if you'd like some help with this, but here's mainly what I was envisioning for this section.

> - Template-driven forms require less boilerplate, setup, and mental overhead

This can play off of the previous section about content projection, mention that you don't need to configure any "back-end" model, you can simply compose the form in your template.
Easy composition of these template-driven forms is facilitated by leveraging content projection so that form logic and state can be handled at the container level.

> Model-driven forms more appropriate for working with complex dynamically generated forms

This might best be expressed through an example, but really the point I was hoping to make is when ngIf's and ngFor's aren't enough, or become unreadable and unmaintainable, then you've probably ventured into territory when model-driven forms are a more appropriate solution.

I don't know if @smithad15 would have anything to add, but I think those are the main points

PR #918, opened 2017-05-26
Copy edit on "10 ways" chapter

gvwilson:
@winkerVSbecks @rpdasilva can you please look this over and comment? In particular, there are a couple of FIXME points (including one entire section where @winkerVSbecks wasn't sure what was wanted).


gvwilson:
See instead #918.

Issue #962, opened 2021-04-18
Update Routing Examples


brenzy:
- [ ] All examples in the Routing chapter are updated to reflect the current version of Angular.  
- [ ] The examples are checked into the [Angular Book Example github repository](https://github.com/rangle/angular-book-examples).
- [ ] The links are updated in the Routing chapter.

Issue #957, opened 2021-04-08
Update - Routing


AbdellaToronto:
## Routing
- [ ] -> should we mention UrlTree?
- [ ] Remove “new” from “new component router”
#### Base URL Tag
- [ ] Mention that generating a project via CLI will do this for you automatically
- [ ] Is the demos sidenote necessary? (if Plunkr might no longer be used)
### Controlling Access to or from a Route
- [ ] In first guard example, remove line 16?
Guard example:
![image](https://user-images.githubusercontent.com/1427502/114886996-0a854380-9dd6-11eb-9285-81990a549eb3.png)

#### Async Route Guards
- [ ] Mention UrlTree also being a possible return value
- [ ] See Official Documentation… link should point to: https://angular.io/guide/router#preventing-unauthorized-access
### Passing Optional Parameters to a Route
- [ ] Update queryParams with queryParamMap
- [ ] See Official Documentation… link should point to: https://angular.io/guide/router#accessing-query-parameters-and-fragments
### Using Auxiliary Routes
- [ ] Review the example and make sure it follows current Angular style

PR #963, opened 2021-04-18
Issue #957: Spring 2021 Routing updates part one.

brenzy:
## Routing
- [X] Should we mention UrlTree? -> Mentioned in routing guard section 
- [X] Remove “new” from “new component router”

#### Base URL Tag
- [X] Mention that generating a project via CLI will do this for you automatically
- [X] Is the demos sidenote necessary? (if Plunkr might no longer be used) -> It doesn't apply to Stackblitz, and the Plunkr demos don't run anyways, so these have all been removed

### Controlling Access to or from a Route
- [ ] In first guard example, remove line 16? -> I think this is referenced elsewhere in the example.  However the whole example should be revisited to reflect the current version of Angular.  Issue #962 added to address updates to the examples.

#### Async Route Guards
- [x] Mention UrlTree also being a possible return value
- [x] See Official Documentation… link should point to: https://angular.io/guide/router#preventing-unauthorized-access

### Passing Optional Parameters to a Route
- [x] Update queryParams with queryParamMap - the example needs some updates in general and will be addressed separately 
- [x] See Official Documentation… link should point to: https://angular.io/guide/router#accessing-query-parameters-and-fragments

### Using Auxiliary Routes
- [] Review the example and make sure it follows current Angular style - this doesn't seem too far off but all examples will be addressed separately.

Issue  #962 - added to address the examples.

I didn't notice any mention of adding routing to a module.  We may want to add an issue for this,

### DIFF
--- a/routing/README.md
+++ b/routing/README.md
@@ -1,4 +1,4 @@
 # Routing
 
-In this section we will discuss the role of routing in Single Page Applications and Angular's new component router.
+In this section we will discuss the role of routing in Single Page Applications and Angular's component router.
 
--- a/routing/aux-routes.md
+++ b/routing/aux-routes.md
@@ -1,6 +1,8 @@
 # Using Auxiliary Routes
 
-Angular supports the concept of auxiliary routes, which allow you to set up and navigate multiple independent routes in a single app. Each component has one primary route and zero or more auxiliary outlets. Auxiliary outlets must have unique name within a component.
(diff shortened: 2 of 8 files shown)


sumitarora (approved):
LGTM but there are still some plnkr examples.

brenzy:
@sumitarora - I will remove all the plunkr examples in a separate issue.

